PR #31, opened 2023-06-16
Basic API client for the Control Room's Asset Storage

### LINKED ISSUE #30: Add Asset Storage library
This MVP will implement the basic client support for interacting with the **Asset Storage** API.

### ToDo
- [x] New `robocorp.storage` package dealing with the assets in a similar way to Work Items.
- [x] Listing, creating, retrieving and deleting assets support.
- [x] Basic [**direct**] asset upload when saving data into the asset.
- [x] Basic documentation.
- [x] Simple real API sweet test skipped from CI.

### ⚠️ This first iteration won't include (destined for follow-up post-release work):
- Updating an asset name and its meta-data. (no upstream API support atm)
- Uploading payloads bigger than 6MB. (requires URL-based asset upload support in the library)
- Uploading assets of other types than text. (like JSON, bytes, files etc.)
- Working locally with assets. (requires the Adapter pattern which mimics the flow with local files)
- Detailed documentation aligned well with the rest of the library.
- Production-ready Portal example show-casing the usage.
- Extended atomic unit-tests with full coverage.
- Legacy `rpaframework` support with RF keywords. (see [comment](https://github.com/robocorp/robo/issues/30#issuecomment-1593726467) below)

cmin764:
Closes #30


storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+__version__ = "0.1.0"
+version_info = [int(x) for x in __version__.split(".")]
+
+LOGGER = logging.getLogger(__name__)
+ASSETS_CLIENT = get_assets_client()
+
+
+def list_assets():
+    response = ASSETS_CLIENT.get("")
+    return response.json()
```
cmin764, line 15: WIP: Type annotations and functions for the rest of the API endpoints.

storage/src/robocorp/storage/_requests.py
```diff
@@ -0,0 +1,180 @@
+import json
```
cmin764, line 1: Next PR: This will be used from a common single place instead of being duplicated.

storage/src/robocorp/storage/_storage.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_assets_client():
+    """Creates and returns an API client based on the injected env vars in CR."""
+    api_url = required_env("RC_API_URL_V1")
+    api_token = required_env("RC_API_TOKEN_V1", "")
+    api_key = required_env("RC_API_KEY", "")  # for local tests only
```
cmin764, line 14: @sahokas Do you think this can pose any security risk? (introducing the ability to make requests with env var injected API keys -- not designed for production, but makes local testing easier)

storage/src/robocorp/storage/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+def url_join(*parts: str) -> Optional[str]:
+    """Join parts of URL and handle missing/duplicate slashes."""
+    if not parts:
+        return None
+
+    url = parts[0].rstrip("/") + "/"
+    for part in parts[1:]:
+        url = urlparse.urljoin(url, part.strip("/") + "/")
+    return url
```
cmin764, line 33: Next PR: This suffered changes I would like to put in the common utilities once we share these between this lib and the **workitems** one.

storage/tests/test_storage.py
```diff
@@ -0,0 +1,8 @@
+from robocorp import storage
+
+
+class TestLibrary:
+
+    def test_list_assets(self):
+        asset_list = storage.list_assets()
```
cmin764, line 7: Simple dummy test that actually works with all the env vars in place. (so the API can be queried)

cmin764 (commented):
Made a self-review to explain myself given the current WIP additions.

storage/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+skip-string-normalization = false
+skip-magic-trailing-comma = false
+
+[tool.isort]
+profile = "black"
+
+[tool.poetry.dependencies]
+python = "^3.9"
+requests = "^2.28.2"
+tenacity = "^8.0.1"
```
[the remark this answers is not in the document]
fabioz: I'm not exactly sure what you have in mind, but `robocorp-tasks` shouldn't really need `requests` nor `tenacity`.

Maybe a better place would be `robocorp.http` for such common bits (and then this one and work items would rely on it?)

storage/tests/test_storage.py
fabioz: As a note, the vault has a test which runs actually integrated with the cloud (with utilities such as getting `rcc` setup from `devutils`)... I guess it'd be nice to have at least one such integration test here -- see: `vault.tests.robocorp_vault_tests.test_vault_api.test_vault_integrated`

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+from ._storage import get_assets_client
+
+
+__version__ = "0.1.0"
+version_info = [int(x) for x in __version__.split(".")]
+
+LOGGER = logging.getLogger(__name__)
+ASSETS_CLIENT = get_assets_client()
```
fabioz, line 10: I'm not sure whether `ASSETS_CLIENT` and `get_assets_client()` should actually be a part of the public API...

storage/pyproject.toml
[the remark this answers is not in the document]
osrjv: We discussed this with @cmin764 ealier today over Zoom, and the point was to have a place for common code that deal with Control Room API requests / environment variables / etc. So mostly work items, asset storage, and process library (at some point).

We don't need to do it immediately, maybe when we need to change the copy-pasted bits it would make sense to harmonize, but it does have a fairly large maintenance overhead. The naming is up in the air, but actually I do like your suggestion of having it also in the public parts of robocorp-http. Then the requests wrapper would actually deliver some proper value.

storage/src/robocorp/storage/__init__.py
osrjv: Probably not. We also discussed earlier today that the client could be lazily initialized using session_cache/task_cache

storage/pyproject.toml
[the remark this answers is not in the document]
fabioz: Maybe it could be used in `robocorp-vault` too... 

I think I saw a few tests failing due to internal errors (500) which could probably be fixed by a retry (https://github.com/robocorp/robo/actions/runs/5231372742/jobs/9445474509)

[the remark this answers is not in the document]
cmin764: Yes, it's about all our Control Room interactions (API requests, env vars handling) bundled together into a common package. We can call it any of the following:
- `robocorp-commons`
- `robocorp-utils`
- `robocorp-core`
- `robocorp-<name>`

What's your suggestion? (ofc, outside of this PR)

storage/src/robocorp/storage/__init__.py
cmin764: Yes and agree! Still, for me, `get_assets_client` is a "protected" API, since it's coming from a private file (*_storage.py*) but is publicly exposed to our own modules using it inside our own package.
And I think this way we bring a lot more clarity to the DX, because:
- *_storage.py* + `get_assets_client()` -> function publicly available for usage inside our own package
- *_storage.py* + `_get_assets_client()` -> function restricted for usage inside its own module only
- *storage.py* + `get_assets_client()` -> function publicly available for any kind usage, especially API interface

Now to make things even more clear, I'd import `get_assets_client as _get_assets_client` so we don't make it "public" accidentally during the package initializer. Will this kind of convention bring more value for you as well?

storage/tests/test_storage.py
cmin764: Good to know, I prepared some complete working tests now (which take 1min to run -- API optimizations incoming) if the required CR env vars (`RC_...`) are in place.

1. Still, since the current `storage` library infrastructure will be refactored once we introduce the Adapter pattern into this library as well, I find it a little bit unnecessary right now, but at least the tests are XFAILED in CI and XPASSED locally with the right env vars supplied.
2. And I think we can improve the state provisioning during the integrated test runs without blending too much the pure testing part with the env setup.

So I'd place 1 & 2 for an upcoming follow-up PR, as on Tuesday `robocorp-storage` should be released so I can already start its wrapping into `rpaframework`.
Do you think we can afford such a minor tech-debt?

cmin764:
Attaching here the archived wheel: 
[robocorp_storage-0.1.0-py3-none-any.whl.zip](https://github.com/robocorp/robo/files/11783743/robocorp_storage-0.1.0-py3-none-any.whl.zip)

storage/pyproject.toml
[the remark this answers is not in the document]
osrjv: I'm a bit hesitant on making a very big "common" library, since it's very hard to keep the API for something like that stable, which makes the maintenance a real pain. The most imporant reusable code (in my mind) is the Control Room API parts, so HTTP requests with retrying and some specific error handling. @fabioz proposed having them as user-facing functionality within `robocorp-http`, which I think could be a good idea, but haven't thought about the specifics.

Things like shared testing code or other non-production things should be inside the `devutils` dependency.

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+from ._requests import Requests, RequestsHTTPError
+from ._storage import AssetNotFound
+from ._storage import get_assets_client as _get_assets_client
+from ._utils import with_lazy_objects as _with_lazy_objects
+
+__version__ = "0.1.0"
+version_info = [int(x) for x in __version__.split(".")]
+
+LOGGER = logging.getLogger(__name__)
+_with_lazy_assets = _with_lazy_objects(_assets_client=_get_assets_client)
```
osrjv, line 15: I get it, but it's a bit of a roundabout way of doing this. You could also `@cache` decorate `get_assets_client` and then call that in functions whenever you need the client, instead of injecting it as arguments.

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@_with_lazy_assets
+def list_assets(*, _assets_client: Requests) -> List[Dict]:
+    """List all the existing assets.
+
+    Returns:
+        A list of assets where each asset is a dictionary with fields like 'id' and
```
osrjv, line 23: Is the shape of the asset defined by the API, or can it be any data that is user-defined? If the former, I'd prefer these to be properly typed, but that can be left as later improvements. Especially if the API is subject to change still.

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Returns:
+        The previously set value of this asset. (empty if it was never set)
+
+    Raises:
+        AssetNotFound: When the queried asset doesn't exist.
+    """
+    LOGGER.info("Retrieving asset %r.", name)
+    payload = _get_asset(name)["payload"]
+    if payload["type"] == "empty":
```
osrjv, line 89: Related to previous comment about types, these would also be easier to understand as a new developer coming in, if the API return types had some shape to them. But still, can be a later improvement.

osrjv (commented):
Some minor comments here and there. Would be good to also add GHA workflows for this

storage/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[tool.poetry.group.dev.dependencies]
+robocorp-devutils = {path = "../devutils/", develop = true}
+types-requests = "^2.30.0.0"
+types-python-dateutil = "^2.8.19.13"
```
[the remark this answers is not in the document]
osrjv: Huh, the above thing about common libraries was meant as a reply, but apparently GH doesn't like replies as part of reviews 😅

storage/pyproject.toml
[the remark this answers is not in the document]
cmin764: Makes sense, thank you! We can discuss more once we decide to refactor the commons into a single source of truth. I'll mark this as "Resolved" since we came to a conclusion.

storage/docs/robocorp.storage.md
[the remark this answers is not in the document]
cmin764: I generated the docs right at the end. I think this is because of the `robocorp.storage` and not `robocorp.storage._storage.py`, right?

storage/src/robocorp/storage/__init__.py
cmin764: Yes, moved to lru-cached function call everytime is needed, best way to do it would be as an adapter property in a class once we translate this into adapter so it can be mirrored by the local files one. Thx!

storage/src/robocorp/storage/__init__.py
cmin764: Yes, defined by the API and quite stable and easy to type and what to expect, good to be done right from this PR.

storage/src/robocorp/storage/__init__.py
cmin764: Yes, totally makes sense here what to expect when you have: `Dict[str, Union[str, Dict[str, str]]]`.

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Args:
+        name: Name of the existing or new asset to create. (if missing)
+        value: The new value set within the asset.
+        wait: Blocks until the new value is reflected within the asset.
+
+    Raises:
+        AssetNotFound: When the queried asset doesn't exist.
+    """
+    existing_asset = _get_asset(name, raise_if_missing=False)
```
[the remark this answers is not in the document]
cmin764: The API is still in the works here and the spec would be that the `name`s are unique and will be directly resolvable by the backend, so we won't need to list all the assets to match the `name` for the `id` anymore.

Maybe this is something that we can clarify with @ovatto / @MarioQuiroga32 if there are critical concerns for this first iteration of the library.

storage/docs/robocorp.storage.md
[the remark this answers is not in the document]
osrjv: Oops, disregard. For some reason I thought it was documenting that file instead of the root of the module.

storage/src/robocorp/storage/__init__.py
osrjv: Yes, I think the singleton / cached client should support adapters later on, and that can be controlled by an env var.

As a sidenote, if we need a way in-code to set the adapter, we can copy the configure pattern from the browser library. So something like:
```python
from robocorp import storage

def some_task():
    storage.configure(adapter=storage.FileAdapter)
    storage.set_asset(...)
```

storage/src/robocorp/storage/__init__.py
osrjv: You can also define more strict types for containers with a guaranteed structure. I think it would be something along these lines:
```python
from typing import TypedDict, Optional, cast

Payload = TypedDict("Payload", {
    "type": str,
    "content_type": Optional[str],
    "url": Optional[str]
})

Asset = TypedDict("Asset", {
    "id": str,
    "name": str,
    "payload": Payload,
})

def get_asset(name) -> Asset:
    return cast(Asset, _client.get_asset(name))
```

storage/src/robocorp/storage/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    if existing_asset:
+        asset_id = existing_asset["id"]
+        LOGGER.debug("Updating already existing asset with ID %r.", asset_id)
+    else:
+        asset_id = _create_asset(name)["id"]
+        LOGGER.debug("Updating newly created asset with ID %r.", asset_id)
+
+    assets_client = _get_assets_client()
+    body = {"content_type": "text/plain"}
+    upload_data = assets_client.post(f"{asset_id}/upload", json=body).json()
```
fabioz, line 124: Question: are there limits on the size of the storage?

If we could query it maybe we could fail earlier (instead of sending all the data and then getting some exception when half-way there).

storage/tests/test_storage.py
```diff
@@ -0,0 +1,55 @@
...
+        assert value == ""
+
+    def test_delete_asset(self, ensured_asset):
+        name = ensured_asset["name"]
+        storage.delete_asset(name)
+        with pytest.raises(storage.AssetNotFound):
+            storage.delete_asset(name)
+
+    def test_set_asset(self):
+        name = "cosmin"
```
fabioz, line 45: We should have tests with weirder unicode ;)

storage/src/robocorp/storage/__init__.py
cmin764: Done and works, thx! Still, we have the minor type ignore when the checker doesn't know if is supposed for the function to raise instead of returning a `None`. So that would avoid:
```
error: Value of type "Optional[Asset]" is not indexable  [index]
```

storage/src/robocorp/storage/__init__.py
cmin764: I think there are some BIG hard limits, but not aware of them. (cc @ovatto )

We would have a 6MB limit if we'd directly send the content to the server in one go, but with the S3 pre-signed ones we're safe to assume "unlimited" for decent sizes.

storage/tests/test_storage.py
```diff
@@ -0,0 +1,55 @@
...
+
+from robocorp import storage
+
+random_str = lambda length: "".join(  # noqa: E731
+    random.choice(string.ascii_lowercase) for _ in range(length)
+)
+
+
+@pytest.mark.xfail(reason="not ready for CI yet (Control Room injected env vars)")
+class TestLibrary:
```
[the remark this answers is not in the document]
cmin764: I copied the model from the legacy lib, where I expect to have other test cases separating concerns like: `TestRobocorpAdapter`, `TestFileAdapter` etc.

But you're on point, this is a YAGNI case, feel free to refactor.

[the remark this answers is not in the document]
fabioz: Thanks for the explanation... I guess I just end up creating different files instead of organizing per classes (and moving common code to a `fixtures` if needed). I guess both work ;)

PR #32, opened 2023-06-19
Tweak logging/docstrings, fix typing issue

osrjv:
The type ignores can be avoided by always raising an AssetNotFound, but catching it internally where relevant. Also did some minor tweaks to docstrings and log lines.

Should not break anything, but couldn't test so not pushing to master.


storage/src/robocorp/storage/__init__.py
```diff
@@ -22,142 +22,128 @@ def list_assets() -> List[AssetMeta]:
...
+            LOGGER.debug("Asset upload still pending, sleeping %.2f...", sleep_time)
+            time.sleep(sleep_time)
+            continue
+        elif status == "completed":
+            break
+        elif status == "failed":
+            reason = upload_data["reason"]
+            raise RuntimeError(f"Asset {name!r} upload failed: {reason!r}")
+        else:
+            raise RuntimeError(f"Asset {name!r} got invalid status: {status!r}")
```
cmin764, line 130: Maybe is even better to have a custom error called `AssetUploadError` deriving from `RuntimeError` (or a better exception class).

storage/src/robocorp/storage/__init__.py
```diff
@@ -22,142 +22,128 @@ def list_assets() -> List[AssetMeta]:
...
-        name: Name of the existing or new asset to create. (if missing)
-        value: The new value set within the asset.
-        wait: Blocks until the new value is reflected within the asset.
+        name: Name of the existing or new asset to create (if missing)
+        value: The new value set within the asset
+        wait: Wait for value to be set succesfully
 
     Raises:
-        AssetNotFound: When the queried asset doesn't exist.
+        RuntimeError: Asset upload failed
```
cmin764, line 97: Good catch! What do you think here of a custom exception, instead of relying on Python's ones? We can check with @MarioQuiroga32 in what cases (like content type reason) an API upload would fail, so we know best from what to derive, `RuntimeError` vs. `ValueError`.

storage/src/robocorp/storage/__init__.py
```diff
@@ -22,142 +22,128 @@ def list_assets() -> List[AssetMeta]:
...
-        asset_id = existing_asset["id"]
-        LOGGER.debug("Updating already existing asset with ID %r.", asset_id)
-    else:
+    try:
+        asset_id = _get_asset(name)["id"]
+        LOGGER.debug("Updating existing asset with id: %r", asset_id)
+    except AssetNotFound:
         asset_id = _create_asset(name)["id"]
-        LOGGER.debug("Updating newly created asset with ID %r.", asset_id)
+        LOGGER.debug("Created new asset with id: %r", asset_id)
```
cmin764, line 104: Looks cleaner, thx!

cmin764 (changes requested):
Good job on finding and fixing those edge-cases, I like where the refactoring is heading!

I'm having a behavior concern with the raise of the `AssetNotFound` in the handler, which will cause retrying and wasting time on sleeps.

storage/src/robocorp/storage/__init__.py
```diff
@@ -22,142 +22,128 @@ def list_assets() -> List[AssetMeta]:
     """
     assets_client = _get_assets_client()
     assets = assets_client.get("").json()
-    LOGGER.info("Found %d assets.", len(assets))
+    LOGGER.debug("Retrieved %d assets", len(assets))
```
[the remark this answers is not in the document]
osrjv: I think without end punctuation, unless there are multiple sentences (and that should be avoided). Although we don't have anything written down for this.

storage/src/robocorp/storage/__init__.py
```diff
@@ -22,142 +22,128 @@ def list_assets() -> List[AssetMeta]:
...
         try:
             assets_client.handle_error(resp)
         except RequestsHTTPError as exc:
             if exc.status_code == 404:
-                exception = exc
+                message = f"Asset {name!r} with id {asset_id!r} not found"
+                raise AssetNotFound(message) from exc
             else:
-                raise
+                raise exc
```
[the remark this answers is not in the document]
osrjv: Should be no difference, I just added it to match the line above so it's maybe a tiny bit more clear what they raise? But I'll take a second look according to other comments

storage/src/robocorp/storage/__init__.py
osrjv: I think RuntimeError (as a base class) is a valid choice since they're not dictated by the user input, but something unexpected from the server. We could make a custom one though, although we'd probably want to bubble that up from all API call errors.

storage/src/robocorp/storage/__init__.py
```diff
@@ -13,157 +16,157 @@
...
+    except RequestsHTTPError as exc:
+        if exc.status_code == 404:
+            # NOTE(cmin764): Any `404 Not Found` error won't be retried and will
+            #  bubble-up from the first request.
+            message = f"Asset {name!r} with id {asset_id!r} not found"
+            raise AssetNotFound(
+                message, status_code=exc.status_code, status_message=exc.status_message
+            ) from exc
+        else:
+            raise exc
```
cmin764, line 32: We use this handler with bot GET and DELETE. (since the asset may not be found in both scenarios)

storage/src/robocorp/storage/__init__.py
```diff
@@ -13,157 +16,157 @@
...
 
     Raises:
-        AssetNotFound: When the queried asset doesn't exist.
+        AssetNotFound: Asset with the given name does not exist
     """
-    LOGGER.info("Deleting asset %r.", name)
+    LOGGER.info("Deleting asset: %r", name)
+    asset_id = _retrieve_asset_id(name)
     assets_client = _get_assets_client()
-    assets_client.delete(_get_asset(name)["id"])  # type: ignore
```
cmin764, line 160: So we'd avoid one extra unnecessary GET.

storage/src/robocorp/storage/_requests.py
```diff
@@ -27,17 +27,18 @@ def _needs_retry(exc: BaseException) -> bool:
     # 400 - payload is bad and needs to be changed
     # 401 - missing auth bearer token
     # 403 - auth is in place, but not allowed (insufficient privileges)
+    # 404 - resource does not exist
```
cmin764, line 30: Since this is planned to get back into a common place and we might need to replicate the behaviour with other resources as well, not just assets.

storage/src/robocorp/storage/_storage.py
```diff
@@ -22,8 +33,12 @@
...
 
 
-class AssetNotFound(KeyError):
-    """Raised when the queried asset couldn't be found."""
+class AssetNotFound(RequestsHTTPError):
+    """No asset with given name/id found."""
+
+
+class AssetUploadFailed(RuntimeError):
+    """There was an unexpected error while uploading an asset."""
```
cmin764, line 41: Let me know if punctuation should be missing on docstrings as well, but according to [PEP8](https://peps.python.org/pep-0008/) these look aligned now.

cmin764 (approved):
LGTM now, thank you!

I tested it thoroughly and added my fixes as well:
- https://github.com/robocorp/robo/pull/32/commits/5e6946206af0991ce82e1ca35a4ebe4c0392ddcc
- https://github.com/robocorp/robo/pull/32/commits/91747f22c2030080ef5d85d37e2fe0a8d8c6b25a

PR #36, opened 2023-06-26
Create metapackage for robocorp libraries

osrjv:
Adds a plain `robocorp` PyPI package that has the core framework features: tasks, logging, and control room libraries.
Has to include _some_ content, so adds a private module `robocorp._meta` with just version information. Could technically include other information that our tooling would read.

Updating of deps is currently `inv update-deps` that makes it match other pyproject.tomls, but this can be further automated if we want.


meta/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+	"Ossi Rajuvaara <ossi@robocorp.com>",
+]
+readme = "README.md"
+repository = "https://github.com/robocorp/robo/"
+license = "Apache-2.0"
+packages = [{include = "robocorp", from="src"}]
+
+[tool.poetry.dependencies]
+python = "^3.9"
+robocorp-tasks = "1.0.0"
```
fabioz, line 15: I think this one should also provide the pinned `robocorp-log` version too (and when we update any version this one would be bumped up).

meta/README.md
```diff
@@ -0,0 +1,5 @@
+# robocorp (metapackage)
+
+A package that includes the core Python libraries of Robocorp: [tasks](https://pypi.org/project/robocorp-tasks/), [logging](https://pypi.org/project/robocorp-log/), and [Control Room](https://robocorp.com/docs/control-room) integration.
```
fabioz, line 3: I think that it could also list the specific libraries here instead of just saying 'control room integration'.

osrjv:
> We should probably have a test which fails when we bump up a version from a dep and don't bump this one.

I think maybe a separate workflow that checks this project against versions in PyPI. There might be times we want to have upgrading this as a manual decision.

> As a note, managing this with poetry is kind of annoying (because one needs to publish the dep i.e.: robocorp-log, wait for it to be live in pypi and then go back to this one update the version and run a poetry lock to then release it -- maybe we can try to automate that somehow or consider some other tool since there's no real code here).

That's true. Could be a traditional setup.py that is generated. I'll fix that later when I get annoyed by this 😅

PR #39, opened 2023-07-04
storage: add json/file support, split APIs, general fixes

### LINKED ISSUE #34: Asset Storage phase 2 improvements
### List of library improvements

- Ability to interact with assets by using `name`s directly. (`name:` prefix endpoint API support required, currently limited to IDs only; benefit: **speed**)
  - Names should be unique.
  - `asset_id` param for instructing the library functions which is an ID and which is the name. (currently, we use both interchangeably)
    - Remove the warning logging of assuming IDs when using names once we provide such param.
  - Suppress the `404` error logging report when we catch the error internally.
- Ability to make direct asset value uploads through the newly introduced `data` param. (saves **time**; limited to **5MB**)
- Upload JSON content type as well. (direct dictionary upload without requiring pre-serialization into text)
- Optional: File-based asset uploads. (setting and retrieving an asset directly from/into a file)
- _\<Add your enhancement here\>_

osrjv:
Fixes #34

### DIFF
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -235,8 +235,7 @@ def set_file(
             content_type = "application/octet-stream"
             LOGGER.info("Unable to detect content type, using %r", content_type)
 
-    with open(path, "rb") as fd:
-        content = fd.read()
+    content = path.read_bytes()
 
     _set_asset(name, content, content_type, wait)
 
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -87,7 +87,7 @@ def _get_asset(name: str) -> "Response":
     else:
         # Note (2023-07-04):
         # The 'data' payload type should only be used when uploading,
-        # and it should never be the the response for getting an asset
+        # and it should never be in the response when getting an asset.
         raise RuntimeError(f"Unsupported payload type: {payload['type']}")
 
 
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -146,8 +146,7 @@ def get_file(name: str, path: Union[os.PathLike, str], exist_ok=False) -> Path:
     if path.exists() and not exist_ok:
         raise FileExistsError(f"File already exists: {path}")
 
-    with open(path, "wb") as fd:
-        fd.write(response.content)
+    path.write_bytes(response.content)
 
     return path
 
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -115,7 +115,7 @@ def get_json(name: str, **kwargs) -> JSON:
         **kwargs: Additional parameters for `json.loads`
 
     Returns:
-        Asset content as a Python object (dict, list, etc.)
+        Asset content as a Python object (dict, list etc.)
 
     Raises:
         AssetNotFound: No asset defined with given name
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -1,172 +1,275 @@
-import functools
+import json
 import logging
-import random
-import time
-from typing import List, cast
+import mimetypes
+import os
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional, Union
 
-import requests
+from ._client import AssetNotFound, AssetUploadFailed
 
-from ._requests import RequestsHTTPError
-from ._storage import Asset, AssetMeta, AssetNotFound, AssetUploadFailed
-from ._storage import get_assets_client as _get_assets_client
+if TYPE_CHECKING:
+    from ._client import AssetsClient
+    from ._requests import Response
 
 __version__ = "0.2.0"
 version_info = [int(x) for x in __version__.split(".")]
 
+JSON = Union[dict[str, "JSON"], list["JSON"], str, int, float, bool, None]
+
 LOGGER = logging.getLogger(__name__)
 
-
-def _handle_missing_asset(response: requests.Response, *, name: str, asset_id: str):
-    assets_client = _get_assets_client()
-    try:
-        assets_client.handle_error(response)
-    except RequestsHTTPError as exc:
-        if exc.status_code == 404:
-            # NOTE(cmin764): Any `404 Not Found` error won't be retried and will
-            #  bubble-up from the first request.
-            message = f"Asset {name!r} with id {asset_id!r} not found"
-            raise AssetNotFound(
-                message, status_code=exc.status_code, status_message=exc.status_message
-            ) from exc
-        else:
-            raise exc
+# Known (additional) mimetypes from file extensions
+KNOWN_MIMETYPES = [
+    ("text/x-yaml", ".yml"),
+    ("text/x-yaml", ".yaml"),
+]
 
 
-def list_assets() -> List[AssetMeta]:
+@lru_cache(maxsize=1)
+def _get_client() -> "AssetsClient":
+    """
+    Creates and returns an Asset Storage API client based on the injected
+    environment variables from Control Room (or RCC).
+    """
+    from ._client import AssetsClient
+    from ._environment import get_endpoint, get_token, get_workspace
+
+    workspace = get_workspace()
+    endpoint = get_endpoint()
+    token = get_token()
+
+    return AssetsClient(workspace, endpoint, token)
+
+
+def list_assets() -> list[str]:
     """List all the existing assets.
 
     Returns:
-        A list of assets where each asset is a dictionary with fields like 'id' and
-        'name'
+        A list of available assets' names
     """
-    assets_client = _get_assets_client()
-    assets = assets_client.get("").json()
-    LOGGER.debug("Retrieved %d assets", len(assets))
-    return assets
-
-
-def _retrieve_asset_id(name: str) -> str:
-    for asset in list_assets():
-        if asset["name"] == name:
-            asset_id = asset["id"]
-            LOGGER.debug("Found existing asset %r with id: %s", name, asset_id)
-            return asset_id
-
-    LOGGER.warning("No asset with name %r found, assuming id", name)
-    return name
-
-
-def _get_asset(name: str) -> Asset:
-    asset_id = _retrieve_asset_id(name)
-
-    assets_client = _get_assets_client()
-    _handle_error = functools.partial(
-        _handle_missing_asset, name=name, asset_id=asset_id
-    )
-    response = assets_client.get(asset_id, _handle_error=_handle_error)
-    return cast(Asset, response.json())
-
-
-def get_asset(name: str) -> str:
-    """Get an asset's value by providing its `name`.
-
-    Args:
-        name: Name of the asset
-
-    Returns:
-        The previously set value of this asset, or empty string if not set
-
-    Raises:
-        AssetNotFound: Asset with the given name does not exist
-    """
-    LOGGER.info("Retrieving asset: %r", name)
-    payload = _get_asset(name)["payload"]
-    if payload["type"] == "empty":
-        LOGGER.warning("Asset %r has no value set", name)
-        return ""
-
-    assets_client = _get_assets_client()
-    url = payload["url"]
-    return assets_client.get(url, headers={}).text
-
-
-def _create_asset(name: str) -> Asset:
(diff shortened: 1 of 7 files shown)


storage/src/robocorp/storage/_client.py
```diff
@@ -0,0 +1,170 @@
...
+        the returned upload URL for delivering the content.
+        """
+        payload = {}
+        if content_type is not None:
+            payload["content_type"] = content_type
+        if data is not None:
+            payload["data"] = data
+
+        path = url_join(quote_plus(asset_id), "upload")
+        handler = self._handle_asset_not_found(asset_id)
```
fabioz, line 162: Is this handler correct? 
i.e.: it seems a bit strange having a `_handle_asset_not_found` when creating an asset.

fabioz (approved):
Nice improvements ;)

storage/src/robocorp/storage/_client.py
fabioz: Resolving as it seems there's a slight possibility of it becoming invalid between the create time/upload time.

storage/src/robocorp/storage/_types.py
```diff
@@ -0,0 +1,104 @@
...
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+
+ContentType = str
+
+Asset = TypedDict(
```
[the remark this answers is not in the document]
osrjv: I'm fairly certain I ran into some issues with some of the fields that are shadowing builtins (id, type), even though I can't for the life of me remember what. This constructor also allows fields that aren't valid Python identifiers, although all the ones here are.

storage/src/robocorp/storage/_client.py
osrjv: Yes, this upload endpoint is for updating the contents of existing assets, and the creation endpoint is separate. The current codepath _does_ always check that it exists before calling this, but I added that handler for every endpoint where 404 means "asset not found", even if unlikely.

storage/src/robocorp/storage/_client.py
```diff
@@ -0,0 +1,170 @@
...
+        self._client.delete(path, _handle_error=handler)
+
+    def upload_asset(
+        self,
+        asset_id: str,
+        content: bytes,
+        content_type: Optional[ContentType] = None,
+        wait=True,
+        poll_interval=0.5,
```
cmin764, line 100:
```suggestion
        wait: bool = True,
        poll_interval: float = 0.5,
```

storage/src/robocorp/storage/_client.py
```diff
@@ -0,0 +1,170 @@
...
+    ):
+        """Upload an asset payload, and optionally wait for it to finish."""
+        if _estimate_base64_size(len(content)) < DATA_LIMIT:
+            LOGGER.info("Content size under data limit, uploading directly")
+            data_content = base64.b64encode(content).decode("ascii")
+            assert len(data_content) < DATA_LIMIT, "Unexpected data size"
```
cmin764, line 106: Instead of asserting, we can just switch to the "full" extensive data upload once we confirm that our assumption was wrong. And the `LOGGER.info("Content size under data limit, uploading directly")` can be placed after this check, so we'd print it once we're sure we're able to upload the data directly, otherwise we fallback to the lengthy upload scenario.

storage/src/robocorp/storage/_types.py
cmin764, line 1: I was using [cattrs](https://pypi.org/project/cattrs/) in some other projects for a very similar purpose. And then the `cast(...)` will refactor in the dataclass instantiation itself.

Not sure if it simplifies anything here, but listing it here just in case you find it useful or more adequate.

storage/src/robocorp/storage/_requests.py
```diff
@@ -58,70 +58,71 @@ def extensive_log(level, msg, *args, **kwargs):
...
+        # Fallback values
+        status_code = response.status_code
+        reason = response.reason
+        message = response.text
+
         try:
             fields = response.json()
+
+            # For some reason we might still get a string from the deserialized
+            # JSON payload, possible due to some double encoding bug in CR
```
cmin764, line 106: Have no idea if this was fixed already, but was happening loooong time ago.

cmin764 (approved):
LGTM and thank you for the huge revamping!

Nothing critical nor bugs, just some nits and things to think about.

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
-from ._storage import Asset, AssetMeta, AssetNotFound, AssetUploadFailed
-from ._storage import get_assets_client as _get_assets_client
+if TYPE_CHECKING:
+    from ._client import AssetsClient
+    from ._requests import Response
 
 __version__ = "0.2.0"
 version_info = [int(x) for x in __version__.split(".")]
 
+JSON = Union[dict[str, "JSON"], list["JSON"], str, int, float, bool, None]
```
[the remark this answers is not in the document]
osrjv: I wonder if this is an RF issue or something else? I know the recursive type support was improved in recent versions of mypy that makes this function correctly

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
-    Args:
-        name: Name of the asset
+    if payload["type"] == "url":
+        url = payload["url"]
+        return Requests().get(url)
+    if payload["type"] == "empty":
+        raise ValueError(
+            f"Asset {details['name']!r} is empty."
+            + " It could mean an upload is still pending"
+            + " or has previously failed."
```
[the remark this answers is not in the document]
osrjv: pyright (and some linters) don't like implicit string concatention and prefer being explicit. Doesn't break anything either way, though, so no hard opinion here.

storage/src/robocorp/storage/_client.py
```diff
@@ -0,0 +1,170 @@
...
+    AssetUploadState,
+    ContentType,
+)
+from ._utils import url_join
+
+LOGGER = logging.getLogger(__name__)
+
+# Assets under `DATA_LIMIT` can be uploaded directly through the POST request,
+# instead of requesting a separate upload url
+DATA_LIMIT = 5 * 10**6  # 5MB
```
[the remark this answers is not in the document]
osrjv: I wasn't sure if CR deals with MB or MiB in this case. I think it would be good to run some tests for this, at least once.

storage/src/robocorp/storage/_types.py
osrjv: It's mostly a built-in way to make type checkers work correctly, it doesn't actually do any runtime verification of the content, as I hope we can rely on CR not changing this API suddenly.

If we want to extend this in any way, I'd prefer using `pydantic`.

storage/src/robocorp/storage/_client.py
osrjv: The calculation _should_ never be wrong, although I guess it can be juggled around to make it behave more nicely if the code is misbehaving in some way

storage/src/robocorp/storage/_requests.py
osrjv: I know, I remember it being a "feature" of some old API. It very much _should_ be fixed in CR instead of having to deal with it in the library.

storage/src/robocorp/storage/_client.py
osrjv: mypy (and other tools) do correct type inference from defaults, but again, I don't have a strong preference here if you think being explicit is better

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
+        **kwargs: Additional parameters for `json.loads`
 
+    Returns:
+        Asset content as a Python object (dict, list, etc.)
 
-def set_asset(name: str, value: str, wait: bool = True):
-    """Creates or updates an asset named `name` with the provided `value`.
+    Raises:
+        AssetNotFound: No asset defined with given name
+        JSONDecodeError: Asset was not valid JSON
```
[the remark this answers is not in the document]
osrjv: I think because we use `.json()` from `requests`, it always guarantees a correct type comes from the response, but I'm not 100% sure because... well... Python exceptions don't have any typing like in Java or other languages.

storage/src/robocorp/storage/__init__.py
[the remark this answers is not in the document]
cmin764: Most probably it was something RF specific if this runs on your side now, so all good!

storage/src/robocorp/storage/_client.py
cmin764: That's interesting, coming from RF where in many cases being explicit was better than implicit and my first reaction was that you forgot to put the types in :).

Might be better to be explicit, which will help with those cases when you come back to the code to edit a parameter and you edit the default into a `None` (or something that isn't trivial for a human to guess, like a value returned by a function call) and you forget to update the type then.
Also being explicit might help with docs generation or IDE auto-completion in those scenarios where the tool is not as smart as **mypy**.

storage/src/robocorp/storage/_client.py
cmin764: Yes, checking the size, and if is small then actually encoding it for the quick transfer check, then checking it again to make sure is really what we expected. And only then decide if we still go on the quick or long way, without raising any error where we're in control.
Would add more peace to us as it makes the code more resilient to such corner case that should never happen.

storage/src/robocorp/storage/_requests.py
cmin764: Yes, can be kept for safety precautions, as it does no harm.

storage/src/robocorp/storage/_client.py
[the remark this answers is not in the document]
cmin764: If I remember correctly, the actual limit is 6MB, but we left some room for the headers (like the lengthy token) and anything else that might add to the payload itself, so actually it doesn't pay the effort to be accurate either way.

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
+    """
+    response = _get_asset(name)
+    return response.json(**kwargs)
 
-    Args:
-        name: Name of the existing or new asset to create (if missing)
-        value: The new value to set within the asset
-        wait: Wait for value to be set successfully
+
+def get_file(name: str, path: Union[os.PathLike, str], exist_ok=False) -> Path:
```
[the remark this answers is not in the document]
osrjv: I used `exist_ok` since that's used in Python standard library (at least in `pathlib`)

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
-        AssetNotFound: Asset with the given name does not exist
+        AssetNotFound: No asset defined with given name
     """
-    LOGGER.info("Retrieving asset: %r", name)
-    payload = _get_asset(name)["payload"]
-    if payload["type"] == "empty":
-        LOGGER.warning("Asset %r has no value set", name)
-        return ""
+    response = _get_asset(name)
+    return response.text
```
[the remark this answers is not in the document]
osrjv: In `requests`, the `content` attribute is the raw decoded bytes (as in UTF-8 text), and `text` is a helper that tries to decode it and also guess the encoding based on the headers it receives from the server

storage/src/robocorp/storage/__init__.py
```diff
@@ -1,172 +1,275 @@
...
+        content = fd.read()
 
+    _set_asset(name, content, content_type, wait)
 
-def delete_asset(name: str):
-    """Delete an asset by providing its `name`.
 
-    Args:
-        name: Name of the asset to delete
+def set_bytes(
```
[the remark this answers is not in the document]
osrjv: The `set_bytes` is the lowest level of the asset storage API. All of the others (text, json, file) can be built on top of it. Most users should use the abstractions, but maybe sometimes they have some binary or document in memory and want to store that.

Maybe it's anticipating a use-case that doesn't exist yet, but the `set_text` approach has the annoying feature that it doesn't play well with the Control Room dashboard

storage/src/robocorp/storage/_client.py
osrjv: Actually, since the limit is actually 6MB, the difference between the estimate and reality is at most a few bytes, I'll just remove the assert. The estimation should be plenty to ensure we're within the "safe limits".

PR #41, opened 2023-07-10
vault: retry support, better error messages

osrjv:
Copy-pasting the `_request.py` here for retry support, which can then next be harmonized into some common package (or similar). Other important change is to propagate proper error messages from responses.


cmin764 (approved):
LGTM!

fabioz (approved):
Looks good. Note: it may be nice to add an entry to `/vault/docs/changelog.md`

PR #43, opened 2023-07-13
Change heuristic to compute whether browser is headless and allow specifying headless through RPA_HEADLESS_MODE.

fabioz:


browser/src/robocorp/browser/_browser_context.py
```diff
@@ -122,24 +122,30 @@ def _browser_config() -> _BrowserConfig:
...
 
 @session_cache
 def browser_type_launch_args() -> Dict:
     launch_options = {}
-    headless = _browser_config().headless
-    if headless is None:
-        launch_options["headless"] = not _is_debugger_attached()
+
+    # RPA_HEADLESS_MODE can be used to force whether running headless.
+    rpa_headless_mode = os.environ.get("RPA_HEADLESS_MODE")
```
osrjv, line 138: It _does_ look like workforce agent sets this in docker runs, which makes this behaviour good as the user can't break the bot by accident. Unless there is some use-case where the user explicitly tries to run headful using xvfb or similar.

I'm slightly annoyed by the fact that it's an envvar with `RPA_` prefix, since we've tried to avoid those in these new libraries in favor of Control Room's `RC_` prefix. I wonder if it would make sense to add a second one that workforce agent supports? `RC_BROWSER_HEADLESS`? Maybe not...

osrjv (approved):
LGTM

browser/src/robocorp/browser/_browser_context.py
fabioz: I guess we should check with the platform guys -- I'm ok with adding it if it'll be used ;)

cmin764: Maybe there was a distinction between `RC_*` (what comes from Control Room / Robocorp) and `RPA_*` (what you set to control the automation, no matter where the bot is running). And sometimes the env var was overriding the explicit parameter, other times the parameter takes precedence.

PR #46, opened 2023-07-18
robcorp.log_pytest: Integration of robocorp.log and pytest.

fabioz:


cmin764 (approved):
LGTM! (although didn't review it in detail)

log_pytest/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+
+# Add the devutils even if the poetry env isn't setup (to do a 'inv devinstall').
+try:
+    import devutils
+except ImportError:
+    devutils_src = Path(__file__).absolute().parent.parent / "devutils" / "src"
+    assert devutils_src.exists(), f"{devutils_src} does not exist."
+    sys.path.append(str(devutils_src))
+
+if True:  # Create block to stop isort from changing the import below.
```
osrjv, line 12: I sort supports comments to control the functionality per-line/block/file: https://pycqa.github.io/isort/docs/configuration/action_comments.html

osrjv (approved):
Looks good, at least from what I can gather. I'll be happy to take this for a spin after release to see how it works. This is not a major priority, but will be cool to show to some testing folks.

fabioz:
> Looks good, at least from what I can gather. I'll be happy to take this for a spin after release to see how it works. This is not a major priority, but will be cool to show to some testing folks.

The one thing I want to add before a release is better support for `assert` statements. After testing it a bit, without it the usage isn't that good for testing (because assertions in general don't have a descriptive message it's hard to know why something actually failed).

PR #48, opened 2023-07-19
Support spaces under asset names

cmin764:
Tests are working with the following exports:
```
RC_API_TOKEN_V1=eyJhbG******S5lcxniwGqjhwUXxZe_7H0qqhvlIa6oXZ
RC_API_URL_V1=https://api.eu1.robocorp.com/v1
RC_WORKSPACE_ID=d6b65aa4-0c45-4fd7-8bec-d68a29896e78
```

Token can be generated with: `rcc cloud authorize -w d6b65aa4-0c45-4fd7-8bec-d68a29896e78`


storage/src/robocorp/storage/_client.py
```diff
@@ -3,7 +3,7 @@
 import math
 import time
 from typing import Optional, cast
-from urllib.parse import quote_plus
+from urllib.parse import quote_plus as sanitize_id
```
cmin764, line 6: So I can easily test with `quote` vs `quote_plus` or other sanitizing functions.

storage/tests/test_storage.py
```diff
@@ -19,37 +19,41 @@ class TestLibrary:
...
+
+    def test_text_asset(self, cleanup_assets):
+        name = "cosmin-text"
+        text = "cosmin@robocorp.com"
+        storage.set_text(name, text)
+        obtained_text = storage.get_text(name)
+        assert obtained_text == text
+
+    def test_bytes_asset(self, cleanup_assets):
+        name = "cosmin - bytes"
```
cmin764, line 55: Currently this fails during later GET retrieval with: `FAILED tests/test_storage.py::TestLibrary::test_bytes_asset - robocorp.storage._client.AssetNotFound: 404 not_found - Asset with id 'name:cosmin - bytes' not found`

osrjv:
It looks like it was an actual CR bug, closed in this PR: https://github.com/robocorp/robocloud-logic/pull/3695

cmin764:
Re-opening it as it improves testing. And I'll also make sure the `+` sign works with the way CR unquotes the URL, but I'll strip away from the PR any unnecessary refactoring.

storage/tests/test_storage.py
cmin764: The API is now ready to deal with spaces, but works with the `%20` encoding for spaces, therefore `quote` instead of `quote_plus` works here now. (cc @osrjv )

PR #49, opened 2023-07-21
Log: When an assert fails more information is provided.

fabioz:


osrjv (commented):
Not the easiest bit of code to review without deep-diving, but looks ok. Having per-package overrides for ruff (based on preferences) is acceptable, as long as there aren't too many ignores.

PR #51, opened 2023-07-21
Provide a way to configure the viewport size / browser context kwargs.

fabioz:


browser/src/robocorp/browser/__init__.py
```diff
@@ -36,20 +36,61 @@ def configure(**kwargs) -> None:
...
+            width, height = value
+            browser_context_kwargs()["viewport"] = {"width": width, "height": height}
+            continue
+
         if not hasattr(config, key):
             raise ValueError(f"Invalid configuration: {key}.")
         setattr(config, key, value)
 
 
+def browser_context_kwargs() -> dict:
```
osrjv, line 65: While this _does_ technically describe what is happening, I feel like we could somehow organize this in some way which is more user friendly. Can we maybe combine creating the context explicitly with this method? Or maybe we can use kwargs in the config function as well?

So instead of this, maybe something like....
```python
def configure_context(**kwargs):
    _browser_context.kwargs = kwargs
```

So then the configuration function matches the actual callsite, instead of being just a dict? A minor difference, but something about this makes me not quite satisfied.

osrjv (commented):
Some thoughts about the new configuration API

browser/src/robocorp/browser/__init__.py
fabioz: I agree that `configure_context` receiving the arguments is better (although I made it `browser_context_kwargs.update(kwargs)` instead of always resetting it).

PR #53, opened 2023-07-24
http: read filename from header, add tests

osrjv:
- Try to read the filename more robustly (from response headers, from url parts) and with a more clear fallback value when _nothing_ matches.
- Removed the `stream` argument as it is not useful to control
- Added some tests too!


fabioz (approved):
LGTM

PR #55, opened 2023-07-26
workitems: allow re-saving output work items

osrjv:


osrjv:
The `saved` flag is kept, because it's used as a way to check if the work item was saved before the execution ends, and notify the user that they probably didn't intend that

cmin764 (approved):
LGTM!

PR #70, opened 2023-10-06
Hackathon: Python logs streaming

cmin764:


cmin764:
Let's keep this for history purposes next time we decide to implement this feature for aiding logs streaming during CR runs.

PR #82, opened 2023-10-24
tasks: add setup/teardown fixtures

osrjv:
Needs some documentation, but here's how it currently works:
```python
from robocorp.tasks import task, setup, teardown


@setup
def before_task(task):
    print(f"Running before task: {task.name}")


@setup(scope="session")
def before_all(tasks):
    names = ", ".join(task.name for task in tasks)
    print(f"Running before tasks: {names}")


@teardown
def after_task(task):
    print(f"Running after task: {task.name}")


@teardown(scope="session")
def after_all(task):
    names = ", ".join(task.name for task in tasks)
    print(f"Running after tasks: {names}")


@task
def first():
    print("Doing things")


@task
def second():
    print("Doing other things")
```

### DIFF
--- a/tasks/docs/guides/02-setups-teardowns.md
+++ b/tasks/docs/guides/02-setups-teardowns.md
@@ -25,7 +25,7 @@ A `Task` object is the internal representation of a parsed task in a Python file
 It contains, for instance, the `filename` and `lineno` of the function, but
 also the `status` and optional `message` after it has been executed.
 
-## Example
+## Example usage
 
 ```python
 from robocorp import browser
@@ -50,5 +50,52 @@ def scrape_website():
 
 @task
 def process_data():
+    ...
+```
+
+## Yielding fixtures
+
+Sometimes it's necessary to access a resource created in a `setup` inside a
+`teardown` fixture. In these cases, it's possible to create a `setup` fixture
+that `yield`s.
+
+The library will call the fixture up until the `yield` statement, execute the task,
+and then call the rest of the fixture (in reverse order):
+
+```python
+import time
+from robocorp.tasks import setup
+
+
+@setup
+def measure_time(task):
+    start = time.time()
+    yield  # Task executes here
+    duration = time.time() - start
+    print(f"Task {task.name} took {duration} seconds")
+
+
+@task
+def my_long_task():
+    ...
+```
+
+It's also possible to access the task's result by assigning a value from
+the `yield` statement:
+
+```python
+from robocorp.import browser
+from robocorp.tasks import setup
+
+
+@setup
+def before_and_after(task_before):
+    print(f"Task before running: {task_before}")
+    task_after = yield
+    print(f"Task after running: {task_after}")
+
+
+@task
+def my_task():
     ...
 ```
--- a/tasks/docs/guides/02-setups-teardowns.md
+++ b/tasks/docs/guides/02-setups-teardowns.md
@@ -1,0 +1,54 @@
+# Setups and teardowns
+
+The library contains two fixture methods, `setup` and `teardown`, that can be
+used to call arbitrary functions before and after tasks are run.
+
+The fixtures receive as arguments the tasks objects that the library has parsed
+from the tasks file, which allows fixtures to behave differently based on which
+task is going to be executed or if task failed or passed.
+
+This is useful for having shared initialization or clean-up steps for all tasks,
+or for ensuring that a function is always executed when implementing libraries.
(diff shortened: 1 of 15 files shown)


fabioz (commented):
+1 one from me. 

The one change I'd do is instead of using `all`, using `session` (since that's the same name already used in session_cache).

Also, I guess this needs some docs (especially to say that the setup/teardown must be explicitly imported to work).

cmin764 (commented):
LGTM! (including Fabio's comment)

fabioz:
I was thinking a bit more and I think that maybe the default should be `tasks`, not `session` if the argument isn't given (as I believe a setup per task is more common than a setup per session).

osrjv:
> I was thinking a bit more and I think that maybe the default should be `tasks`, not `session` if the argument isn't given (as I believe a setup per task is more common than a setup per session -- that also follows what `pytest` does, so, pytest users may be expecting that).

Makes sense. Would also make the argument of `ITask` instead of `Sequence[ITask]` a bit more obvious and easy to use. I think in 99% of cases we want users to only run one task at once anyways.

osrjv:
Review fixes:
- Changed naming to `task` and `session`
- Changed `task` to be default
- Added method docs and a guide

Will still change the setup callbacks to raise on error

fabioz:
I was actually thinking that most of my use-cases are actually a setup along with a teardown (so, writing a case where there's setup/yield/teardown would be nice).

osrjv:
> I was actually thinking that most of my use-cases are actually a setup along with a teardown (so, writing a case where there's setup/yield/teardown would be nice).

How do you see that from a usage and naming PoV? It's neither a setup nor a teardown, but more of a generic `fixture` like in pytest. I guess I could add an additional function that supports it, similar to the caching decorators

fabioz:
I think `fixture` is ok, but even `setup` accepting a `yield` inside it would work for me...

osrjv:
> I think `fixture` is ok, but even `setup` accepting a `yield` inside it would work for me...

I went with this. Here's what a yielding setup looks like:
```python
@setup
def measure_time(task):
    start = time.time()
    yield  # Task executes here
    duration = time.time() - start
    print(f"Task {task.name} took {duration} seconds")
```

Added also relevant documentation

cmin764 (commented):
Wondering what happens if you decorate with both a `setup` which yields and a `teardown`. Would both the below yielding part and the other teardown logic execute? (haven't tested it yet)

osrjv:
> Wondering what happens if you decorate with both a `setup` which yields and a `teardown`. Would both the below yielding part and the other teardown logic execute? (haven't tested it yet)

Yes, there is no limit to the amount of setup and teardown fixtures you can register. Setups are executed in registered order and teardowns in reverse registered order.

With this code:
```python
from robocorp.tasks import setup, teardown, task

@setup
def yielding(task):
    print("Generator before yield")
    yield
    print("Generator after yield")

@teardown
def first(task):
    print("Teardown #1")

@teardown
def second(task):
    print("Teardown #2")

@task
def a_task():
    print("Inside task")
```

The following would be printed:
```
Generator before yield
Inside task
Generator after yield
Teardown #2
Teardown #1
```

tasks/docs/guides/02-setups-teardowns.md
```diff
@@ -0,0 +1,101 @@
...
+
+```python
+from robocorp.import browser
+from robocorp.tasks import setup
+
+
+@setup
+def before_and_after(task_before):
+    print(f"Task before running: {task_before}")
+    task_after = yield
```
fabioz, line 94: This got me a bit puzzled. Isn't it so that both `task_before` and `task_after` are the same object?

i.e:  `assert task_before is task_after` should work.

If this is true, I'm not sure whether returning the task from yield does make much sense (not that it couldn't be done, but it's not very useful given that it'd just return the task in the parameter).

osrjv: Good catch. I just noticed the same and had it already removed locally.

It would maybe be better if we didn't pass the mutable task object to the user, but maybe that has some use-cases.

fabioz (approved):
LGTM!

tasks/docs/guides/02-setups-teardowns.md
osrjv: We might have to change the inner workings a bit to actually support passing a run by changing the status of the task objects, I think? I'll try it out, actually

PR #85, opened 2023-10-26
tasks: allow changing run state from fixtures

osrjv:
Some minor fixes to `tasks` that makes the final run status (and exit code) depend on the values stored in task objects. Which in turn allows affecting them in the new teardown fixtures.

### DIFF
--- a/tasks/src/robocorp/tasks/_commands.py
+++ b/tasks/src/robocorp/tasks/_commands.py
@@ -3,7 +3,7 @@ import os
 import sys
 import traceback
 from pathlib import Path
-from typing import List, Optional, Sequence, Union
+from typing import List, Literal, Optional, Sequence, Union
 
 from ._argdispatch import arg_dispatch as _arg_dispatch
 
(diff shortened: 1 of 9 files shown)


fabioz:
LGTM, so, +1 from me (missing a unit-test for it though...).

osrjv:
Added a small test for mutating the state.

Now that the task status/state is part of the user-facing API, I changed it to a proper string enum, and separated it from the status of the logging. Take a quick look

tasks/src/robocorp/tasks/_commands.py
```diff
@@ -229,16 +228,17 @@ def run(
...
-        run_status = "PASS"
-        setup_message = ""
-
         run_name = os.path.basename(p)
         if task_name:
             run_name += f" - {task_name}"
 
+        # Status string from `log` module
+        # TODO: Replace with enum
+        run_status: Union[Literal["PASS"], Literal["ERROR"]] = "PASS"
```
fabioz, line 237: Question: is there a reason you went with the union of literals here and not just the Status enum directly? Does something break when doing that?

osrjv: ... although I guess I could use the task status enum here as well, and then just convert when calling log methods, but not sure if really more readable

fabioz: Ok, agreed.

fabioz:
+1 -- ship it ;)

PR #87, opened 2023-10-26
Don't automatically redact common python words such as None, True, Fa…

fabioz:
…lse as well as strings with less than 3 characters.


osrjv:
Sorry if this apparent from the code, but is it possible to redact these if configured explicitly?

PR #88, opened 2023-10-30
Add `robocorp-controlroom` for operating with the new Control Room API

cmin764:
### Blockers
- ~Fixing the pydantic types: `HasMore`, `Next`.~ ✅ 

### ToDo
- [ ] Remove the _research_ when not needed anymore.
- [ ] Remove the tests directory. (as it's empty and not needed)
- [ ] Generate code in a subdir called `_client`. (so we isolate it from the rest of the public-interfaced libs)
- [ ] Auto-pass `workspace_id` by patching the generated API class.
- [ ] Fix README on reference: `[API Key with permissions](../README.md#API Key with permissions)`
- [ ] Handle request errors & retry like happens in the other libs.
- [ ] Ensure `truststore` injection and TLS behaviour still works.

### DIFF
--- a/workspace/bin/post-process-module.py
+++ b/workspace/bin/post-process-module.py
@@ -1,8 +1,16 @@
 #! /usr/bin/env python
 
+
+import re
 import sys
 from functools import lru_cache
 from pathlib import Path
+
+
+RE_MODEL_TO_PRIMITIVE = re.compile(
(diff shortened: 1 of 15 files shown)


joshyorko:
This is over 2 years old folks,  lets close this out.

PR #90, opened 2023-11-03
Fix assets listing and support pagination & limit control

### LINKED ISSUE #89: `storage`: `list_assets` fails to parse response due to changed structure
Test error: `FAILED tests/test_storage.py::TestLibrary::test_list_assets - TypeError: string indices must be integers`
Cause: JSON structure changed into putting the assets list into `data` field:
```
{'has_more': False, 'next': None, 'data': [{'id': '7399aa63-34a8-4b8d-98b5-864b8a3cab30', 'name': 'cosmin-wzzlyqrx'}, {'id': '94d7dd92-2a26-4aa7-8653-91fdfe26bfe1', 'name': 'mentions_black_list'}, {'id': '9cf6373e-7803-4a87-9fcf-e2f2cdaef3bc', 'name': 'test-ossi-malformed'}, {'id': 'c8cb6e7d-ea5d-4cbe-8c8a-fa8b8ae838db', 'name': 'websites_to_monitor'}, {'id': 'f7e91925-6df2-4291-b7c7-453a91a2820c', 'name': 'test'}]}
```

Additionally `has_more` and `next` have been added, thus we need to cover these when requesting and also control the number of returned assets from the server through the `limit` parameter. ([API](https://robocorp.com/api/asset/create-asset-upload))

cmin764:
Closes #89 

### Main fixes
- Assets resources retrieval

### Others
- Add pagination control (~with optional `page_limit`~)
- Add `mdformat` dev dep required on Mac for building docs with `inv docs`
- Removed dev command "set-version-in-deps"

### DIFF
--- a/devutils/pyproject.toml
+++ b/devutils/pyproject.toml
@@ -28,9 +28,6 @@ tomlkit = "^0.11.8"
 semver = "^3.0.0"
 pylint = "^2.17.4"
 
-[tool.poetry.group.dev.dependencies]
-mdformat = "^0.7.17"
-
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
--- a/devutils/pyproject.toml
+++ b/devutils/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-devutils"
-version = "0.0.1"
+version = "0.0.2"
 description = "Utilities for development in monorepo."
 authors = [
 	"Fabio Z. <fabio@robocorp.com>",
@@ -27,6 +27,9 @@ lazydocs = { git = "https://github.com/robocorp/lazydocs.git" }
 tomlkit = "^0.11.8"
 semver = "^3.0.0"
 pylint = "^2.17.4"
+
+[tool.poetry.group.dev.dependencies]
+mdformat = "^0.7.17"
 
 [build-system]
 requires = ["poetry-core"]
--- a/devutils/pyproject.toml
+++ b/devutils/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-devutils"
-version = "0.0.2"
+version = "0.0.1"
 description = "Utilities for development in monorepo."
 authors = [
 	"Fabio Z. <fabio@robocorp.com>",
--- a/devutils/src/devutils/invoke_utils.py
+++ b/devutils/src/devutils/invoke_utils.py
@@ -321,15 +321,6 @@ def build_common_tasks(root: Path, package_name: str, tag_prefix: Optional[str] 
             sys.exit(1)
 
     @task
-    def set_version_in_deps(ctx, version):
-        """Sets a new version of this project in the project dependencies"""
-        root_pyproject = root / "pyproject.toml"
-        assert root_pyproject.exists(), f"Expected {root_pyproject} to exist."
-
-        all_pyprojects = list(collect_deps_pyprojects(root_pyproject))
-        all_pyprojects.append(root_pyproject)
-
-    @task
     def set_version(ctx, version):
         """Sets a new version for the project in all the needed files"""
         import re
--- a/storage/docs/CHANGELOG.md
+++ b/storage/docs/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fix `list_assets` response parsing and support pagination size control with
+  the `page_limit` parameter.
+
 ## 1.0.1 - 2023-10-13
 
 - Fix support for Python 3.8
--- a/storage/docs/CHANGELOG.md
+++ b/storage/docs/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-- Fix `list_assets` response parsing and support pagination size control with
-  the `page_limit` parameter.
+- Fix `list_assets` response parsing.
 
 ## 1.0.1 - 2023-10-13
 
--- a/storage/docs/CHANGELOG.md
+++ b/storage/docs/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 1.0.2 - 2023-11-06
+
 - Fix `list_assets` response parsing.
 
 ## 1.0.1 - 2023-10-13
--- a/storage/pyproject.toml
+++ b/storage/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-storage"
-version = "1.0.1"
+version = "1.2.0"
 description = "Robocorp Asset Storage library"
 authors = [
 	"Cosmin P. <cosmin@robocorp.com>",
--- a/storage/pyproject.toml
+++ b/storage/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-storage"
-version = "1.2.0"
+version = "1.1.0"
 description = "Robocorp Asset Storage library"
 authors = [
 	"Cosmin P. <cosmin@robocorp.com>",
--- a/storage/pyproject.toml
+++ b/storage/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-storage"
-version = "1.1.0"
+version = "1.0.2"
 description = "Robocorp Asset Storage library"
 authors = [
 	"Cosmin P. <cosmin@robocorp.com>",
--- a/storage/src/robocorp/storage/__init__.py
+++ b/storage/src/robocorp/storage/__init__.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
     from ._client import AssetsClient
     from ._requests import Response
 
-__version__ = "1.1.0"
+__version__ = "1.0.2"
 version_info = [int(x) for x in __version__.split(".")]
 
 JSON = Union[Dict[str, "JSON"], List["JSON"], str, int, float, bool, None]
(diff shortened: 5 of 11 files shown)


devutils/pyproject.toml
```diff
@@ -28,6 +28,9 @@ tomlkit = "^0.11.8"
 semver = "^3.0.0"
 pylint = "^2.17.4"
 
+[tool.poetry.group.dev.dependencies]
+mdformat = "^0.7.17"
```
cmin764, line 32: This was required when I ran for the first time `inv docs` in **storage**. (placed here in case it is required by other packages as well)

```bash
% inv docs
poetry run lazydocs --validate --no-watermark --remove-package-prefix --src-base-url https://github.com/robocorp/robo/tree/master/ --overview-file README.md --output-path /Users/cmin/Repos/robo/storage/docs/api robocorp.storage
Traceback (most recent call last):
  File "/Users/cmin/Repos/robo/storage/.venv/bin/lazydocs", line 5, in <module>
    from lazydocs._cli import app
  File "/Users/cmin/Repos/robo/storage/.venv/lib/python3.9/site-packages/lazydocs/__init__.py", line 6, in <module>
    from .generation import MarkdownGenerator, generate_docs
  File "/Users/cmin/Repos/robo/storage/.venv/lib/python3.9/site-packages/lazydocs/generation.py", line 18, in <module>
    import mdformat
ModuleNotFoundError: No module named 'mdformat'
```

devutils/src/devutils/invoke_utils.py
```diff
@@ -320,15 +320,6 @@ def check_tag_version(ctx):
...
 
-    @task
-    def set_version_in_deps(ctx, version):
-        """Sets a new version of this project in the project dependencies"""
-        root_pyproject = root / "pyproject.toml"
-        assert root_pyproject.exists(), f"Expected {root_pyproject} to exist."
-
-        all_pyprojects = list(collect_deps_pyprojects(root_pyproject))
-        all_pyprojects.append(root_pyproject)
-
```
cmin764, line 331: Not needed anymore. (Slack [thread](https://robocorptechnologies.slack.com/archives/C05CQR7JLNT/p1687343211057589?thread_ts=1687302523.420139&cid=C05CQR7JLNT))

storage/docs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fix `list_assets` response parsing and support pagination size control with
+  the `page_limit` parameter.
+
```
cmin764, line 7: Will make it into a release post-merge, so the release tag gets on top of it.

storage/src/robocorp/storage/__init__.py
```diff
@@ -42,13 +42,36 @@ def _get_client() -> "AssetsClient":
     return AssetsClient(workspace, endpoint, token)
 
 
-def list_assets() -> List[str]:
+def list_assets(page_limit: Optional[int] = 100) -> List[str]:
```
cmin764, line 45: Is this something we'd want to still retrieve all the existing assets, or do we want to retrieve the next `page_limit` ones for every subsequent manual call?

Example:
1. `list_assets(5)` -> first 5 items
2. `list_assets(10)` -> next 10 items

If yes, we can keep a context of the upcoming `next_page`, although quite useless and non-testable since the API isn't ready yet.

storage/src/robocorp/storage/__init__.py
```diff
@@ -42,13 +42,36 @@ def _get_client() -> "AssetsClient":
...
     """List all the existing assets.
 
+    Args:
+        page_limit: How many assets to retrieve per request
```
cmin764, line 49:
```suggestion
        page_limit: How many assets to retrieve per requested page (increase it if your asset isn't returned)
```
Just in case we want better rewording.

storage/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-storage"
-version = "1.0.1"
+version = "1.2.0"
```
fabioz, line 3: Why 1.2 and not 1.1?

storage/src/robocorp/storage/_client.py
```diff
@@ -68,9 +68,22 @@ def _handle_error(response: Response):
 
         return _handle_error
 
-    def list_assets(self) -> List[Asset]:
+    def list_assets(
+        self, *, limit: Optional[int], next_page: Optional[str]
```
fabioz, line 72: `next_page` in this API seems strange (I get that in the previous API it returned what'd be the next page, but here this is the `current_page` right -- the param should probably be just `page` I guess).

storage/src/robocorp/storage/_client.py
```diff
@@ -68,9 +68,22 @@ def _handle_error(response: Response):
...
+                "Choose between a limit (for the initial request) or the next page URL"
+                " (for any subsequent request)"
+            )
+        params = None
+        url = ""
+        if limit:
+            params = {"limit": limit}
+        elif next_page:
+            url = next_page
+        return self._client.get(url, params=params).json()
```
cmin764, line 86: If there's no `url` nor `limit` passed, a default API limit is implicitly used when requesting all the available assets for the first time.

Otherwise we start with a `limit` and as long as we get a `next_page` we'll continue to retrieve assets from that one.

cmin764 (commented):
Self-review

storage/pyproject.toml
cmin764: Good catch!

storage/src/robocorp/storage/_client.py
cmin764: Yes, `page` perfectly inlines with `limit`.

Where on the API side with have `next_page` and `page_limit` so we understand what is this about.

devutils/pyproject.toml
fabioz: But then this shouldn't be here... Rather it should be in the `devutils` `pyproject.toml`.

cmin764: Hmm, I moved it actually to [devutils/pyproject.toml](https://github.com/robocorp/robo/pull/90/files/3ab309cdcc31b476011b3007a740388125eed50b#diff-74077cadcb57aa19e6ddb0eb0cf03cd645d31154a89061abb758130015c76f97) isn't that? (initially I placed it in **storage**'s one, then refactored before push)

storage/src/robocorp/storage/__init__.py
fabioz: I was about to comment that this is a strange. If this will return all assets, then what's the point of the page_limit?

cmin764: Without it you have no control over how many assets will be returned by the API, nor you can increase such a limit. Also it's still not clear to me how the API will [respond](https://robocorptechnologies.slack.com/archives/C053DV6NDPA/p1699004750884589?thread_ts=1699001644.429899&cid=C053DV6NDPA) in the future. (cc @bendersej )

osrjv: I think exposing the pagination could be left entirely out of this version, and rethink when it becomes an actual problem or the API is properly implemented. And even then, I'd say that should be part of the low-level client libraries you're working on and this should be more high level.

devutils/pyproject.toml
fabioz: Seems I saw it wrong there, sorry for the noise.

devutils/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-devutils"
-version = "0.0.1"
+version = "0.0.2"
```
osrjv, line 3: bumping this is not really needed since we don't want to publish ever. you just need to `poetry lock` in the depending package

devutils/pyproject.toml
osrjv, line 32: I'm surprised if this actually fixed things for you. I'm fairly certain this should be in the normal dependencies section to actually get it into use in the downstream package. Poetry _shouldn't_ include dev dependencies from a dependency

osrjv: Just like pylint, black, etc. are currently defined

storage/src/robocorp/storage/__init__.py
```diff
@@ -48,7 +48,27 @@ def list_assets() -> List[str]:
     Returns:
         A list of available assets' names
     """
-    return [asset["name"] for asset in _get_client().list_assets()]
+    client = _get_client()
+    page = client.list_assets()
```
osrjv, line 52: More of a thought than anything else, but if the new APIs prefer having pagination in these types of endpoints, the upcoming API client libraries probably should have some nice abstraction for dealing with paginated responses.

devutils/pyproject.toml
cmin764: Correct for normal package installations, but storage among others as well will install it in dev mode because of this `robocorp-devutils = { path = "../devutils/", develop = true }`, which attracts the dev ones as well. (and kept this as a dev dep since it's for dev only, when we generate the docs)

Let me know if is best to move it somewhere else.

storage/src/robocorp/storage/__init__.py
cmin764: Yes, that will be a very interesting task for `robocorp.workspace` given the raw Py API clients.

devutils/pyproject.toml
cmin764: Actually, the best place for `mdformat` should be as a normal dep in [lazydocs](https://github.com/robocorp/lazydocs.git), isn't it?

osrjv: Everything under devutils is development time only, as the name implies. Not a big deal, but all the other tools are are "normal" dependencies.

Actually, what I'm most confused about is why you don't get mdformat. It should come as a dependency for lazydocs (via mdformat-gfm), and I don't need this fix to generate docs for storage.

cmin764: I was getting it, my bad, forgot to run `poetry update` first before running the first `inv docs`.

PR #93, opened 2023-11-09
[WIP] Add local task server

osrjv:
This is a work-in-progress branch for adding a local task server. Requires further work before release, and is entirely missing the actual React front-end implementation. However, introduces some concepts and skeleton for building the project. Here's a rundown of things:

**Tasks library**
- Parse input and output schema for tasks from function signature
- Expose schema in `list` command
- Allow passing task arguments after `--` with `run` command, e.g. `python -m robocorp.tasks run tasks.py -- --arg "value"`

**Server library**
- Introduce new Python package: `robocorp-task-server`
- Allows starting local HTTP server as a barebones Control Room, e.g. Control Coffin
- Runs inside same environment as project, uses `robocorp-tasks` for collecting tasks and running them
- Bundles the front-end as a static file inside the package and serves it
- Stores project-locally a history of tasks and runs (in sqlite3)
- Offers an API for inspecting and running tasks (`http://localhost:8080/docs`)
- Can watch project for changes and re-collect tasks
- Allows starting tasks with arguments, validates payload based on schema given by tasks library
- Serves local artifacts statically, available at `http://<addr>:<port>/artifacts/<run-id>/<artifact-name>`

**CLI**
- Add `serve` command for starting local task server
- Probe capability of environment (tasks, local server) before running commands, add result to environment cache
- Change configuration file from `pyproject.toml` to new `tasks.toml`. The actual format is not final, and not all fields are used yet, but this is the first proposal.
- Add configuration for "tasks file", e.g. `tasks.py`
- Add base for new template that uses server library

**Next steps**
- Go through all `TODO` comments
- Nail down new `tasks.toml` spec
- Figure out if running tasks in subprocess is the only way to go, or if calling directly via Python API is feasible
- Figure out how to pass task output values to server (related to above)
- Figure out how task arguments / returns map to work items concept, can these be integrated so that a work item maps into the arguments automatically?
- Add log output streaming, can use same approach as VSCode if a websocket endpoint is added that is injected to runs
- Add robo CLI support for `run` arguments passing
- Generate client bindings for backend with OpenAPI generator. Schema available with `python -m robocorp.task_server --schema openapi.json`
- **Implement entire front-end** 😅


cmin764:
Hey @fabioz, what do you think it will work best for getting a quick UI that can use the API? (React (and use our DS), Bootstrap, _\<something else\>_)

fabioz:
I'll go with react / DS (since we use it in multiple places already).

fabioz:
As a note, I'm closing this PR. The idea will be first brewing the task server in a different repo (and moving to the mainline when we agree it's what we want).

I'll extract the change to add arguments to tasks into a separate PR (and we can still reference what we need from the closed PR).

PR #96, opened 2023-12-04
feat: add possibility to expose the server to internet

ktaube:
Adds `--expose` flag to server `start` command.

- Establishes a connection to `robocorp.link`
- Proxies all received calls to the local server
- Try to reconnect on connection loss

## TODO

- [x] More testing. Especially the retry logic.


rihardsgravis (approved):
LGTM 👏

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -0,0 +1,113 @@
...
+        return client.post(payload.path, json=payload.body)
+    elif payload.method == "PUT":
+        return client.put(payload.path, json=payload.body)
+    elif payload.method == "DELETE":
+        return client.delete(payload.path)
+    else:
+        raise NotImplementedError(f"Method {payload.method} not implemented")
+
+
+async def expose_server(app: FastAPI):
```
fabioz: Hummm... hadn't we agreed that this server forwarding would be in its own process?

In this case you'd create a new file and call it with `subprocess.call([sys.executable, ...])`.


Note: it should pass its pid to the child so that the child can automatically exit when the parent process exits -- the code I have around to do that is below (you can add it to the `robocorp.action_server._robo_utils.process` module):

```
_track_pids_to_exit = set()
_watching_thread_global = None


def exit_when_pid_exists(pid):
    _track_pids_to_exit.add(pid)
    global _watching_thread_global
    if _watching_thread_global is None:
        import time

        def watch_parent_process():
            # exit when any of the ids we're tracking exit.
            while True:
                for pid in _track_pids_to_exit:
                    if not is_process_alive(pid):
                        # Note: just exit since the parent process already
                        # exited.
                        log.info(
                            f"Force-quit process: %s because parent: %s exited",
                            os.getpid(),
                            pid,
                        )
                        os._exit(0)

                time.sleep(PARENT_PROCESS_WATCH_INTERVAL)

        _watching_thread_global = threading.Thread(target=watch_parent_process, args=())
        _watching_thread_global.daemon = True
        _watching_thread_global.start()
```

Also, then you can use `requests` instead of `httpx` in that case so that the new dep isn't needed.

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -0,0 +1,113 @@
...
+                            continue
+                        except Exception:
+                            pass
+
+                        try:
+                            payload = BodyPayload(**data)
+                            # might be a bit hacky, but works elegantly
+                            client = TestClient(
+                                app,
+                                base_url=f"http://{settings.address}:{settings.port}",
```
fabioz: The port needs to be queried from fast api because the port could be == 0 in the settings passed by the user (in which case a port would be chosen by the system).

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -0,0 +1,113 @@
...
+                        data = json.loads(message)
+
+                        try:
+                            session_payload = SessionPayload(**data)
+                            log.info(
+                                f"🌍 URL: https://{session_payload.sessionId}.{settings.expose_url}/openapi.json"
+                            )
+                            continue
+                        except Exception:
+                            pass
```
fabioz, line 87: Maybe it'd be nice to log the exception?

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -0,0 +1,113 @@
...
+                ) as ws:
+                    while True:
+                        message = await ws.recv()
+
+                        data = json.loads(message)
+
+                        try:
+                            session_payload = SessionPayload(**data)
+                            log.info(
+                                f"🌍 URL: https://{session_payload.sessionId}.{settings.expose_url}/openapi.json"
```
rihardsgravis: Updated the proxy server to resolve main path to `/openapi.json`:
```suggestion
                                f"🌍 URL: https://{session_payload.sessionId}.{settings.expose_url}"
```

action_server/src/robocorp/action_server/_server_expose.py
ktaube: ok, refactoring.

But what are we gaining with separate process? Why not just run it in the same process / event loop?

ktaube: Also, `subprocess.call` is a blocking call, so using `subprocess.Popen` instead

action_server/src/robocorp/action_server/_server_expose.py
ktaube: This can fail if e.g. the payload is `BodyPayload`.
I added extra check to raise exception if the initial session creation fails.

action_server/src/robocorp/action_server/_server_expose.py
ktaube: This is now running in a separate process. Not sure how to get this from the parent process.
Maybe let's just enforce explicit port setting?

action_server/src/robocorp/action_server/_server_expose.py
fabioz: >But what are we gaining with separate process? Why not just run it in the same process / event loop?

The benefit is working with actual concurrency.

fabioz: I agree maybe the benefit is not that big (but I also think that the change would be small...).

action_server/src/robocorp/action_server/_server_expose.py
fabioz: You can get it from fastapi and pass it in the command line parameters.

action_server/src/robocorp/action_server/_server_expose.py
ktaube: This is done. please review!

fabioz: This part seems good. Resolving conversation.

action_server/src/robocorp/action_server/_server_expose.py
ktaube: Done 👍

action_server/src/robocorp/action_server/_server.py
```diff
@@ -104,5 +101,38 @@ async def serve_index(request: Request):
...
         )
 
+    def check_port(loop):
+        if not server.started:
+            loop.call_later(1 / 15.0, partial(check_port, loop))
+            return
+
+        port = settings.port if settings.port != 0 else None
+        if port is None:
+            port = server.servers[0].sockets[0].getsockname()[1]
```
fabioz: There's a caveat here. You could have more than one socket, one ipv4 and one ipv6. In this case, you need to do a for and check which one is ipv4 (because the _server_expose.py can't deal with ipv6 in the current structure).

ktaube: https://github.com/robocorp/robo/pull/96/commits/9f24e8c3e52bfdfa2f094af9991f813cc2bb4d20 should do the trick 👍

PR #99, opened 2023-12-07
Feat: Client application trace header support

rihardsgravis:
Adds an  Action Server feature to pass an `X-action-trace` header to the action run endpoint  and log it at action preload step.

Note: Added the FastAPI header signature so that additional headers could be added in future, but not sure if the headers themselves should be defined inside the function or would there be a better place for this?


fabioz (approved):
Looks good to me. Feel free to merge once the tests finish running ;)

PR #100, opened 2023-12-11
Release `windows` 1.0.0

### LINKED ISSUE #98: `windows`: Generic improvements
A few observations to tackle before releasing **v1**.

- Documentation:
  - Be specific about having the `timeout` in seconds under the docstrings
  - Fix `calc = find_window("name:Calculator")` into `calc = windows.find_window("name:Calculator")` in the main docs
  - API docs: not all the arguments are formatted properly as a bullet list (maybe due to inconsistent spacing, like leading EOL)
- Behaviour:
  - List user only processes (when matching a window based on PID) for speed gain
  - `windows.find()` might be too verbose in the terminal output (even printing the element tree multiple times when there are no matches) -> can be configured to be less verbose
  - Fix unicode error post-run during console message write (see trace below) -> encoding issue to be solved in `tasks` (might be fixed by setting UTF-8 as encoding through env var)
  - Searching (most probably wrongly) for a control with `Desktop().find()` will lead into a lot of trace console output (until the timeout depletes) compared to the recommended pattern of `windows.find_window().find()` -> bad usage pattern
- Functionality:
  - A locator like `'name:"Page setup" type:WindowControl'` doesn't find the "Notepad" > "Page setup" window with `windows.find_window()` but it works with `RPA.Windows.control_window()` (example below) -> designed like this on purpose for better & faster experience
  - `.find()` and `.find_window()` fails and raises when something isn't found, what about logging a warning and returning `None` if a parameter like `strict=False` is set explicitly? (so we don't need to guard with try-except every time we test the existence of a control/window)

Test [Task Package](https://github.com/cmin764/robots/tree/main/kari-notepad)

----

Screenshot 2023-12-06 at 1 42 59 PM

```
Error calling: <bound method Context._after_task_run of <robocorp.tasks._task.Context object at 0x000001E754521D80>>.
Traceback (most recent call last):
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_callback.py", line 48, in __call__
    c(*args, **kwargs)
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_task.py", line 169, in _after_task_run
    show(f"{msg}", kind=status_kind)
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_task.py", line 142, in show
    console_message(msg, kind, flush=flush)
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\log\__init__.py", line 374, in console_message
    stream.write(message)
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\log\redirect.py", line 114, in write
    self._original.write(s)
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\ue711' in position 6207: character maps to <undefined>
```

```python
from RPA.Windows import Windows
from robocorp import windows
from robocorp.tasks import task

WIN = Windows()

@task
def notepad_automation():
    page_setup_locator = 'name:"Page setup" type:WindowControl'
    WIN.control_window(page_setup_locator)
    page_window = windows.find_window(page_setup_locator)
```

```
...
Thread 0x00003050 (most recent call first):
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\threading.py", line 320 in wait
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\queue.py", line 171 in get
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_log_output_setup.py", line 73 in write_in_thread
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\threading.py", line 953 in run
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\threading.py", line 1016 in _bootstrap_inner 
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\threading.py", line 973 in _bootstrap        

Current thread 0x000018ac (most recent call first):
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\windows\_vendored\uiautomation\uiautomation.py", line 6350 in GetFirstChildControl
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\windows\_vendored\uiautomation\uiautomation.py", line 6403 in GetChildren
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\windows\_iter_tree.py", line 142 in iter_tree
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\windows\_control_element.py", line 582 in _iter_children_nodes
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\windows\_control_element.py", line 487 in find
  File "z:\Repos\robots\kari-notepad\tasks.py", line 34 in notepad_automation
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_task.py", line 30 in run
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_commands.py", line 367 in run
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_argdispatch.py", line 20 in _dispatch
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\_argdispatch.py", line 167 in process_args
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\cli.py", line 41 in main
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\site-packages\robocorp\tasks\__main__.py", line 4 in <module>
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\runpy.py", line 86 in _run_code
  File "C:\ProgramData\robocorp\ht\b395beb_b1f3c24_9ba66f27\lib\runpy.py", line 196 in _run_module_as_main   
Windows fatal exception: code 0x80040155
...
```

cmin764:
Closes #98 

### ToDo
- [x] Introduce the `raise_error` param in `find()`, when set to `None` won't raise but log and return null instead.
- [x] Apply the `@overload` typing decoration so it produces the right type annotations for the non-raise case.
- [x] Apply the same pattern for the `find_window` ~and `find_windows`~ function~s~. -> `find_windows` should rely on `wait_for_window=False` to not raise any error and return what was currently found
- [x] ~When finding windows based on PID filter from the user's processes only. (not all as it takes time)~ -> not necessary, as we don't iterate over all the existing processes, but obtain the executable name of the retrieved process ID of the controlled element
- [x] Unit-test coverage.
- [x] Documentation fixes: docstrings, library visibility in README & docs site.
- [x] Major release.

### Next
- Portal examples with the latest release.
- New Inspector integration.
- Related Issue with local tests run: https://github.com/robocorp/robo/issues/105


windows/src/robocorp/windows/_control_element.py
```diff
@@ -431,11 +431,50 @@ def fmt(s):
...
 
+    def _raise_or_warn_control_not_found(
+            self,
+            locator: Locator,
+            search_depth: int,
+            timeout: Optional[float],
+            raise_error: Optional[type],  # to raise or not and with what class
+            exception: Exception,  # previous original raised exception
+    ):
+        from . import config as windows_config
```
fabioz, line 441: I think we shouldn't generate the verbose version of the error when `raise_error` is False (as that's slow and this use-case says that the error is expected...).

cmin764: Totally agree, and isn't this already reflected in the current [change](https://github.com/robocorp/robo/pull/100/files#diff-b3a3f67db1c482a1b6c0953faf82ccd4a8d47acdf50c1cf127e8b62d69416c4fR461-R464)? (the long verbose message won't be generated with this set as `None`)

Also, I went with `raise_error` for non-ambiguity (clearer than `default`), but this will take either `None` or the exception type to be raised along with the message.

fabioz: Oh, yes, sorry :)

windows/src/robocorp/windows/_control_element.py
```diff
@@ -431,11 +431,50 @@ def fmt(s):
...
+            child_elements_msg.append(str(child))
+        msg += "\n".join(child_elements_msg)
+        raise raise_error(msg) from exception
+
     def find(
         self,
         locator: Locator,
         search_depth: int = 8,
         timeout: Optional[float] = None,
+        raise_error: Optional[type] = ElementNotFound,
```
fabioz: I didn't really understand why `raise_error`is a type and not just a boolean...

cmin764: Fixed!

windows/src/robocorp/windows/__init__.py
```diff
@@ -114,13 +114,50 @@ def config() -> "Config":
...
+    timeout: Optional[float] = ...,
+    wait_time: Optional[float] = ...,
+    foreground: bool = ...,
+    raise_error: Literal[False] = ...,
+) -> Optional["WindowElement"]:
+    ...
+
+
+@overload
+def find_window(
```
fabioz, line 142: I'm missing an overload where the `raise_error` is not passed and then it always returns `WindowElement`... or does it infer that the default is `True` and then does the right decision based on the other overloads?

p.s.: if it's the case that it's needed it'd need to be changed in other places too.

windows/src/robocorp/windows/__init__.py
```diff
@@ -114,13 +114,50 @@ def config() -> "Config":
...
+    locator: Locator,
+    search_depth: int = ...,
+    timeout: Optional[float] = ...,
+    wait_time: Optional[float] = ...,
+    foreground: bool = ...,
+    raise_error: bool = ...,
```
cmin764, line 148: The `@overload` needs to happen for the variable type value (`bool`) as well to please mypy.

windows/src/robocorp/windows/__init__.py
```diff
@@ -130,20 +167,20 @@ def find_window(
...
-        timeout:
-            The search for a child with the given locator will be retried
-            until the given timeout elapses.
-
+        timeout: The search for a child with the given locator will be retried
+            until the given timeout (in **seconds**) elapses.
             At least one full search up to the given depth will always be done
             and the timeout will only take place afterwards.
-
             If not given the global config timeout will be used.
```
cmin764, line 174: No empty lines allowed due to formatting issues of the parameters.

cmin764 (commented):
Self-review

windows/src/robocorp/windows/__init__.py
cmin764: When is not passed explicitly, it is inferred from the fact that it is assumed to be "passed" with `True` implicitly. At least **mypy** doesn't scream in this case. Let me know if you think otherwise and if there's any trouble with the actual change, as I'm trying to not add unnecessary code.

PR #103, opened 2023-12-11
feat: add `--expose-session` to restart server and get the same expose URL

ktaube:
Prints id that can be passed to `--expose-session` command to get the same expose server URL.
Useful when e.g. using with GPTs.

## Preview

https://github.com/robocorp/robo/assets/1479451/e33ed410-f7eb-4791-ab26-7c901df16d8e

### DIFF
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -74,7 +74,6 @@ async def expose_server(
         session_payload: Optional[SessionPayload] = (
             get_expose_session_payload(expose_session) if expose_session else None
         )
-        print(session_payload)
         while retries < max_retries:
             try:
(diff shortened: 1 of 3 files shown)


action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -26,6 +27,19 @@ class BodyPayload(BaseModel):
     headers: dict
 
 
+def get_expose_session(payload: SessionPayload) -> str:
+    return f"{payload.sessionId}:{payload.sessionSecret}".encode("ascii").hex()
```
ktaube, line 31: I started with passing both session secret and id as separate args, but IMHO encoding them in a single token makes for better ergonomics.

rihardsgravis (approved):
👍

PR #107, opened 2023-12-12
Chore: Format Client Application Trace as an HTML anchor element

rihardsgravis:
Format Client Application Trace as an HTML anchor element

image


fabioz (approved):
LGTM!

PR #113, opened 2023-12-21
feat: consequential actions

ktaube:
Pass `@action(is_consequential=True)` to add [`x-openai-isConequential`](https://platform.openai.com/docs/actions/consequential-flag) flag to openapi spec.

- Pass arbitrary options to `@task(option="this is option")` decorator that can be accessed through e.g. `task.options.get("option")`
- Pass `@action(is_consequential=True)` option to tell client that the action should be approved by the user

## Preview

```python
from datetime import datetime

from robocorp.actions import action


@action(is_consequential=True)
def get_current_date() -> str:
    """
    Returns the current date in ISO format.
    """
    return datetime.now().isoformat()
```

Former action will now have additional `x-openai-isConsequential` option in the openspec which the GPTs will interpret.

```json

{
  "openapi": "3.1.0",
  "info": { "title": "Robocorp Actions Server", "version": "0.1.0" },
  "paths": {
    "/api/actions/actions-robo/get-current-date/run": {
      "post": {
        "summary": "Get Current Date",
        "description": "Returns the current date in ISO format.",
        "operationId": "get_current_date",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": { "$ref": "#/components/schemas/GetCurrentDateInput" }
            }
          },
          "required": true
        },
        "responses": {
          "200": {
            "description": "Successful Response",
            "content": {
              "application/json": {
                "schema": {
                  "type": "string",
                  "title": "Response Get Current Date"
                }
              }
            }
          },
          "422": {
            "description": "Validation Error",
            "content": {
              "application/json": {
                "schema": { "$ref": "#/components/schemas/HTTPValidationError" }
              }
            }
          }
        },
        "x-openai-isConsequential": true
      }
    }
  },
  ...
}

```


action_server/pyproject.toml
```diff
@@ -26,13 +24,14 @@ uvicorn = "^0.23.2"
 pydantic = "^2.4.2"
 jsonschema = "^4.19.2"
 PyYAML = "^6"
-robocorp-actions = "^0.0.2"
+# robocorp-actions = "^0.0.2"
+robocorp-actions = { path = "../actions", develop = true }
```
ktaube: wip: release new package version and replace

action_server/src/robocorp/action_server/_actions_import.py
```diff
@@ -189,7 +189,11 @@ def _add_actions_to_db(
             "Expected it to be 0.0.2 or higher"
         )
 
-    cmdline = [env.get("PYTHON_EXE", sys.executable), "-m", "robocorp.actions", "list"]
+    # FIXME: PYTHON_EXE will use robocorp.actions from holotree env.
+    #        Why? IMHO, we need to use the same python here as the action-server is using.
+    # python = env.get("PYTHON_EXE", sys.executable)
```
ktaube: @fabioz why are we using python from holotree here, not the same one the action-server is useing? Bug or I'm missing something?

actions/pyproject.toml
```diff
@@ -2,20 +2,19 @@
...
 repository = "https://github.com/robocorp/robo/"
 license = "Apache-2.0"
-packages = [{include = "robocorp", from="src"}]
+packages = [{ include = "robocorp", from = "src" }]
 
 [tool.poetry.dependencies]
 python = "^3.9"
-robocorp-tasks = "^2.6.0"
+# robocorp-tasks = "^2.6.0"
+robocorp-tasks = { path = "../tasks", develop = true }
```
ktaube: wip: release new package version and replace

ktaube: @fabioz is there a simple way to always use the local packages when developing locally and versioned ones only when releasing?

action_server/pyproject.toml
fabioz: This is not needed (and should be reverted).

We have a command: `inv devinstall` which will temporarily do this change to install the env in devmode and then it'll switch back to the proper env aferwards.

action_server/src/robocorp/action_server/_actions_import.py
fabioz: You're missing something. 

We can't use the python from the action server, we *must* use the python being used to run the target robot.

Consider the following: the user has a conda that depends on numpy and has a numpy import. In the action server we don't have the numpy dependency, so, if we try to use our python to load it we'll crash.

The proper way to develop here is create an environment, do `inv devinstall` (which will get the development version in the environment for all libraries) and develop based on that env.

fabioz: i.e.: this change must be reverted.

actions/pyproject.toml
fabioz: Yes: `inv devinstall` (please revert these changes).

tasks/pyproject.toml
```diff
@@ -23,6 +23,7 @@ python = "^3.9"
 robocorp-log = ">=2.4,<3"
 psutil = "^5.9"
 docstring_parser_fork = "^0.0.5"
+pydantic = "^2.5.2"
```
fabioz: Humm, at the tasks level I don't think we should use pydantic, just use a "plain" dataclass for the use-case if needed.

tasks/src/robocorp/tasks/_commands.py
```diff
@@ -60,6 +60,7 @@ def list_tasks(path: str, glob: Optional[str] = None) -> int:
                     "docs": getattr(task.method, "__doc__") or "",
                     "input_schema": task.input_schema,
                     "output_schema": task.output_schema,
+                    "options": task.options.model_dump(),
```
fabioz: Note: with a dataclass this would be `asdict`.

fabioz (changes requested):
Please make sure you create tests for this change. I suggest doing it while developing and not afterwards so that it helps to drive the development (but it must be there anyways).

It should have at least:

- 1 test in the `tasks` -- you can follow just add an additional test in `tasks_tests.test_task_schema` (following the same structure at the `test_task_schema` method)

- 1 integration test in the `action_server`, verifying that the openapi.json is correct in this case. It should follow the same thing that `action_server_tests.test_server.test_import_no_conda` does.

tasks/pyproject.toml
ktaube: sure, will revert.

I just prefer to use validation models in cases like this (e.g. zod in node land). We get
1. automatic data validation with nice errors.
2. typings out of the box

fabioz: As a note, when using dataclasses:
- type-validation does work already with mypy (so, static analysis is ok)
- typings are out of the box too

It doesn't do the runtime type-checking though. That's a tradeoff (so, it's faster), but on (the few) cases where you are having user input you need to validate manually.

actions/src/robocorp/actions/__init__.py
```diff
@@ -32,11 +33,17 @@ def enter_user() -> str:
...
 
-    # i.e.: This is just a thin layer for the task decorator at this point
-    # (it may be extended in the future...).
-    from robocorp.tasks import task
+    def decorator(func, **kwargs):
+        # i.e.: This is just a thin layer for the task decorator at this point
+        # (it may be extended in the future...).
+        from robocorp.tasks import task
 
-    return task(func)
```
fabioz, line 39: Question: can't you just `return task(*args, **kwargs)` here?

actions/src/robocorp/actions/__init__.py
```diff
@@ -8,7 +9,7 @@
 version_info = [int(x) for x in __version__.split(".")]
 
 
-def action(func):
+def action(*args, **kwargs):
```
fabioz, line 25: For the `action` (not `task`), we should add the `@overload`s with the signatures which are valid (see: https://github.com/robocorp/robo/blob/master/windows/src/robocorp/windows/_window_element.py#L136 as an example).

tasks/src/robocorp/tasks/_schemas.py
```diff
@@ -0,0 +1,6 @@
+from pydantic import BaseModel
+from typing import Optional
+
+
+class TaskOptions(BaseModel):
+    is_consequential: Optional[bool] = False
```
fabioz: Why is this `Optional[bool]` and not just `bool`?

action_server/poetry.lock
```diff
@@ -1173,123 +1189,126 @@ name = "robocorp-tasks"
...
 [package.dependencies]
-docstring_parser_fork = ">=0.0.5,<0.0.6"
-psutil = ">=5.9,<6.0"
+docstring_parser_fork = "^0.0.5"
+psutil = "^5.9"
+pydantic = "^2.5.2"
 robocorp-log = ">=2.4,<3"
 
+[package.source]
+type = "directory"
```
fabioz: This lock file should probably not be added (the lock needs to be done based on the pypi deps, not on the devinstall deps).

actions/src/robocorp/actions/__init__.py
fabioz: Humm, this was marked as resolved, but I'm not sure it was done...

I meant for the full implementation of action to be:
```
def action(*args, **kwargs):
    return task(*args, **kwargs)
```

Doesn't that work for some reason?

tasks/src/robocorp/tasks/_task_options.py
```diff
@@ -0,0 +1,6 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class TaskOptions:
```
fabioz: Actually, I was thinking, this isn't correct at this level. 

For the `task`, `is_consequential` does not make sense, just in the `actions`.

Thus, in the tasks we just want the options to be stored as a regular dict with the options the user entered (no typing or checks done as to what the options are).

actions/src/robocorp/actions/__init__.py
```diff
@@ -8,7 +9,12 @@
...
 
-def action(func):
+@overload
+def action(func: Callable, is_consequental: bool = False) -> Callable:
+    ...
+
+
+def action(*args, **kwargs):
     """
     Decorator for actions (entry points) which can be executed by `robocorp.actions`.
```
fabioz, line 27: The `actions` docstring should be updated to show docs for the `is_consequential`.

Also, please add an entry do `docs/changelog.md` outlining this change.

tasks/src/robocorp/tasks/_task_options.py
fabioz: p.s.: at the `task` level, the feature is that `options` are now available to be passed in the decorator (and clients can then introspect those at `task.options` if they want to handle them) -- please add an entry to the tasks/docs/changelog.md outlining this change.

ktaube: Should be done

1. removed `TaskOptions`
2. args passed to `@task` will be accessible through `task.options`
3. added `ActionOptions` dataclass

actions/src/robocorp/actions/__init__.py
ktaube: Don't we want to validate the `@action(...)` args?
Currently, I'm checking `kwargs` to be valid `ActionOptions` so we need some extra logic.

mkorpela:
`x-openai-isConsequential=false` has also special meaning:
"If the x-openai-isConsequential field is false, we show the "always allow button"."
https://platform.openai.com/docs/actions/consequential-flag

Might be good to default to `None` and then have ability to set the flag false and true.

ktaube:
> `x-openai-isConsequential=false` has also special meaning: "If the x-openai-isConsequential field is false, we show the "always allow button"." https://platform.openai.com/docs/actions/consequential-flag
> 
> Might be good to default to `None` and then have ability to set the flag false and true.

Good catch. I made it so:
1. `@action` -> will not add the option
2. `@action(isConsequential=True)` -> will add the `x-openai-isConsequential: true`
3. `@action(isConsequential=False)` -> will add the `x-openai-isConsequential: false`

fabioz (approved):
LGTM!

PR #115, opened 2024-01-04
DRAFT:Changing meta-package deps. to major version ranges to avoid collisions

kariharju:
Problem: At the moment we cannot have `robocorp` and `robocorp-actions` in the same environment due to depencency collision:

```
The conflict is caused by:
    robocorp 1.2.4 depends on robocorp-tasks==2.4.2
    robocorp-actions 0.0.2 depends on robocorp-tasks<3.0.0 and >=2.6.0
```

Options:
- The meta-package deps are changed to semver major ranges, so that other packages can pin more specifically
- We keep the meta-package pinned, which means we need to keep meta-package and other robocorp- packages inline in deps.

Draft because:
- Needs some invoke running to get the poetry.lock up dated..?
- Also updating the authors for the package


fabioz:
It's missing the `poetry lock` command to update poetry deps according to the new range.

I'm not sure. The downside here (especially with the rcc caches) is:
- User installs robo==1.2.5 and gets a set of dependencies
- We release a new robocorp-log bugfix/feature release (no breaking changes)
- robo is not released since it already covers that version given the current range

Result: user never gets the new robocorp-log bugfix because rcc caches are always bound to the versions used when robo==1.2.5 was released (unless he manually pins robocorp-log, which kind of defeats the purpose of the robo meta-package).

kariharju:
> Result: user never gets the new robocorp-log bugfix because rcc caches are always bound to the versions used when robo==1.2.5 was released (unless he manually pins robocorp-log, which kind of defeats the purpose of the robo meta-package).

True, this is the problem with the meta-packages...
Basically whenever we update a sub-package, we really should bump up the meta-package version as well.
With versions locked to specifics like now using the meta-package basically prevents us from using any other library that requires different versions.. or we need to keep all the sub-packages and the meta-package aligned with the packages that are not in the meta-package.

fabioz:
Well, if other libraries don't pin specific internal versions it shouldn't be a problem in general (i.e.: having just major versions is a problem for the meta-package, but it's not a problem for other packages which depend on the internal versions).

fabioz:
I was thinking a bit more and I think that this is reasonable (then at new bugfix versions we can bump the minors here and do a new release of the meta package). 

As a note, the `robo/meta/tasks.py` (used in invoke) has some utilities to deal with those that probably should be updated to deal with that syntax (i.e.: `inv update` may need to be updated to deal with the `^` for versions now).

kariharju:
Done in other branches

PR #116, opened 2024-01-04
feat: persist expose session

ktaube:
Save the expose session to `{cwd}/.robocorp/expose_session.json`.

## Preview

https://github.com/robocorp/robo/assets/1479451/0954cbbb-af5a-4b35-8085-7ac8e840475f

### DIFF
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -48,7 +48,8 @@ def read_expose_session_json(datadir: str) -> None | ExposeSessionJson:
         pass
     except ValidationError as e:
         log.error("Failed to load previous expose session", e)
-        pass
+    except json.JSONDecodeError as e:
+        log.error("Failed to decode exopse session json", e)
     return session_json
 
 
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -36,6 +36,7 @@ class ExposeSessionJson(BaseModel):
 def get_expose_session_path() -> str:
     return os.path.join(os.getcwd(), ".robocorp", "expose_session.json")
 
+
 def read_expose_session_json() -> None | ExposeSessionJson:
     session_json = None
     try:
@@ -56,7 +57,11 @@ def write_expose_session_json(expose_session: str, url: str) -> None:
         log.debug(f"🗂️ Creating .robocorp directory path={dir_path}")
         os.makedirs(os.path.dirname(expose_session_path))
     with open(expose_session_path, "w") as f:
-        json.dump(ExposeSessionJson(expose_session=expose_session, url=url).model_dump(), f, indent=2)
+        json.dump(
+            ExposeSessionJson(expose_session=expose_session, url=url).model_dump(),
+            f,
+            indent=2,
+        )
 
 
 def get_expose_session(payload: SessionPayload) -> str:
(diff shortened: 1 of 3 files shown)


action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -27,6 +28,42 @@ class BodyPayload(BaseModel):
     headers: dict
 
 
+class ExposeSessionJson(BaseModel):
+    expose_session: str
+    url: str
+
+
+def get_expose_session_path() -> str:
+    return os.path.join(os.getcwd(), ".robocorp", "expose_session.json")
```
fabioz: Humm, instead of persisting in the current dir, it could be persisted in the settings dir, since that dir is already tied to the current directory (i.e.: `get_settings().datadir`), that way the user doesn't get its current directory dirty due to that.

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -27,6 +28,42 @@ class BodyPayload(BaseModel):
...
+
+
+def read_expose_session_json() -> None | ExposeSessionJson:
+    session_json = None
+    try:
+        expose_session_path = get_expose_session_path()
+        log.debug(f"🗂️ Reading expose_session.json path={expose_session_path}")
+        with open(expose_session_path, "r") as f:
+            session_json = ExposeSessionJson(**json.load(f))
+    except FileNotFoundError:
```
fabioz, line 47: I think it should also handle json loading errors and issues creating the ExposeSessionJson constructor (i.e.: in case the format changes in the future we don't want errors here in case the user for some reason uses 2 different versions in the same directory).

action_server/src/robocorp/action_server/cli.py
```diff
@@ -114,10 +115,10 @@ def _create_parser():
         help="Expose the server to the world",
     )
     start_parser.add_argument(
-        "--expose-session",
-        dest="expose_session",
-        help="Restart action server with existing expose session",
-        default=None,
+        "--expose-y",
```
kariharju: Naming is hard... could this be:
`--expose-allow-reuse` or something.

action_server/src/robocorp/action_server/_server_expose.py
kariharju: This is actually a really good point. 
Having some way to handle possible changes in the JSON as we can expect things to develop and change at a rapid rate.

ktaube: Added `ValidationError` handler that logs the error.
We're using pydantic model for validating the input, so should be easy to handle changes. We can just set proper defaults for updated fields or implement a new schema for new data shape.

action_server/src/robocorp/action_server/_server_expose.py
ktaube: Moved the `expose_session.json` to `settings.datadir`.
Using `get_settings` in expose loop threw "not initialized" error though, so I just changed it to pass `datadir` in process args.

action_server/src/robocorp/action_server/cli.py
ktaube: Renamed to `--expose-allow-reuse` 👍

action_server/src/robocorp/action_server/_server_expose.py
fabioz: @ktaube I think it should also handle JSONDecodeError (i.e.: if by any reason the file becomes corrupted I believe it can be silently ignored instead of showing an exception to the user).

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -27,6 +28,41 @@ class BodyPayload(BaseModel):
...
+    try:
+        expose_session_path = get_expose_session_path(datadir)
+        log.debug(f"🗂️ Reading expose_session.json path={expose_session_path}")
+        with open(expose_session_path, "r") as f:
+            session_json = ExposeSessionJson(**json.load(f))
+    except FileNotFoundError:
+        pass
+    except ValidationError as e:
+        log.error("Failed to load previous expose session", e)
+        pass
```
fabioz: Minor: as you already have a statement here the `pass` is redundant.

fabioz (approved):
LGTM!

PR #117, opened 2024-01-04
Adding GHA workflow for action server executable signing and S3 publish

kariharju:
TBD:
- Sync the artifact names and workflow name of the workflow that produces the unsigned executables
- The build workflow must also provide a version.txt artifact
- Terraforming on S3 is not yet done
- Testing


.github/workflows/action_server_binary_release.yml
```diff
@@ -70,4 +75,38 @@ jobs:
...
+        if: ${{ matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-11'  }}
+        with:
+          name: ${{ runner.os }}-action-server-wrapper
+          path: "action_server/go-wrapper/action-server-wrapper"
+
+      - uses: actions/upload-artifact@v4
+        if: matrix.os == 'windows-2019'
+        with:
+          name: ${{ runner.os }}-action-server-wrapper
+          path: "action_server/go-wrapper/action-server-wrapper.exe"
```
kariharju: The output executable needs to be named `action-server` as that is what users will be writing in CLI.

action_server/go-wrapper/main.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		fmt.Println("Error reading version.txt:", err)
+		os.Exit(1)
+	}
+	version := strings.TrimSpace(string(versionData))
+
+	// Determine the appropriate path based on the operating system
+	switch runtime.GOOS {
+	case "windows":
+		appDataDir := os.Getenv("APPDATA")
```
kariharju: Needs to be LOCALAPPDATA.
APPDATA points to `C:\Users\<user>\AppData\Roaming\robocorp` and we do not need this to be roaming.

PR #118, opened 2024-01-05
Replacing the template behind `action-server new`

kariharju:
To enable working with current VS Code features


fabioz:
Just merged.

PR #122, opened 2024-01-09
chore: reduce rcc output when running `action-server`

ktaube:
Reduce RCC logs. Show them with `--verbose` option.

Also, print holotree python interpretor path so it can be more easily found for setting up VSCode. Not everyone will want to use Robocorp VSCode extension.

### DIFF
--- a/action_server/src/robocorp/action_server/_actions_import.py
+++ b/action_server/src/robocorp/action_server/_actions_import.py
@@ -120,7 +120,7 @@ Note: no virtual environment will be used for the imported actions, they'll be r
     if not name:
         name = import_path.name
 
-    log.info(f"Python Interpretor path: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}/bin/python")
+    log.info(f"Python Interpreter path: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}/bin/python")
 
     action_package = ActionPackage(
         id=action_package_id,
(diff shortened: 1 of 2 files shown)


rihardsgravis (approved):
💥

action_server/src/robocorp/action_server/_actions_import.py
```diff
@@ -120,6 +120,8 @@ def import_action_package(
     if not name:
         name = import_path.name
 
+    log.info(f"Python Interpretor path: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}/bin/python")
```
rihardsgravis: Typo
```suggestion
    log.info(f"Python Interpreter path: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}/bin/python")
```

action_server/src/robocorp/action_server/_actions_import.py
```diff
@@ -120,6 +120,10 @@ def import_action_package(
     if not name:
         name = import_path.name
 
+    log.info(
+        f"Python Interpretor path: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}/bin/python"
```
kariharju: This only works for linux and mac.
cross-platform way would be to state something like:
`f"Action Package Python environment at: {use_env.get('RCC_HOLOTREE_SPACE_ROOT')}"`

kariharju (approved):

PR #133, opened 2024-01-11
feat: reuse same api key for restarted expose sessions

ktaube:
Saves `api_key` to `ExposeSessionJson` so we can restart the expose session with the same API key.

Can still be overwritten by passing `--api-key <key>` option.

### DIFF
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -31,6 +31,7 @@ class BodyPayload(BaseModel):
 class ExposeSessionJson(BaseModel):
     expose_session: str
     url: str
+    api_key: str | None = None
 
 
 def get_expose_session_path(datadir: str) -> str:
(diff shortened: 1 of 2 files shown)


rihardsgravis (approved):

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -53,12 +54,14 @@ def read_expose_session_json(datadir: str) -> None | ExposeSessionJson:
     return session_json
 
 
-def write_expose_session_json(datadir: str, expose_session: str, url: str) -> None:
+def write_expose_session_json(
+    datadir: str, expose_session_json=ExposeSessionJson
```
fabioz, line 58: I think you meant to write:
`expose_session_json:ExposeSessionJson`
instead of:
`expose_session_json=ExposeSessionJson`

i.e.: set the type, not the value

fabioz: Just saw that it's already merged. I'll fix it here then.

ktaube: ah, you're correct. Thanks!

PR #134, opened 2024-01-11
Fix: [Action Server] Fix Action run result UI crash

rihardsgravis:
Fix UI crash where run result was expected as a string, but may be multiple types


action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -177,7 +180,7 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
           <Header size="small">
             <Header.Title title="Result" />
           </Header>
-          <Code lineNumbers={false} value={result.errorMessage || (result.data as string)} />
+          <Code lineNumbers={false} value={result.errorMessage || stringifyResult(result.data)} />
```
ktaube, line 183: Could be done by builtins maybe
```suggestion
          <Code lineNumbers={false} value={result.errorMessage || String(result.data ?? '')} />
```

ktaube (approved):

PR #141, opened 2024-01-11
Remove `http` package in favor to `requests`

cmin764:
Remove code/docs which mislead ReMark, but also leave room for such 3rd-party recommendation visibility.


kariharju:
We need to make the deprecation release before merging this one.

cmin764:
> We need to make the deprecation release before merging this one.

We can still make the release afterwards by re-running this [job](https://github.com/robocorp/robocorp/actions/runs/7493097462/job/20435409660), as this will use the `0.4.1` tag. So these events can happen async.

PR #154, opened 2024-01-15
feat: ping pong heartbeats

ktaube:
Send `ping` to tunnel every 2 seconds. Run retry logic if there's no answer for 2 seconds.

Tunnel changes: https://github.com/robocorp/action-server-tunnel/pull/4


action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -107,119 +131,176 @@ async def expose_server(
...
+                        log.info("Reconnected to expose tunnel server")
+                        retries = 0
+
+                    ping_task = asyncio.create_task(
+                        handle_ping_pong(ws, pong_queue, no_connection_queue)
+                    )
+
+                    try:
+                        while True:
+                            done, pending = await asyncio.wait(
```
ktaube, line 170: Heartbeats are running in a separate loop, so I'm using the `no_connection_queue` to pass a signal that we have lost connection to the tunnel.

kariharju:
2s heartbeat may be a bit much, between Control Room and Worker we send once a minute.
There's probably a higher detection rate needed here, but maybe reduce 5-10s..?

fabioz (commented):
Not sure, the logic seems quite tricky (and I think it could be simplified).

i.e.: I think that the separate task shouldn't be needed... it should be possible to send a ping and then wait for the return of the ping or any other message with a timeout in the original loop.

Also, the additional task will keep sending pings but they won't be checked afterwards for the pong because the `pong` is only checked at the beginning (of the loop)... is this the expected behavior or should it be verified during other commands too (such as during a run command)?

So, not approving nor disapproving at this time. Waiting for 2nd opinion from @rihardsgravis or @mkorpela.

ktaube:
Yeah, it could be simplified. But currently, this works and would fix the re-connection issues. Not sure I can rework this before the release.

rihardsgravis:
I have no strong opinion, no experience on Python side for this. Heartbeats working in parallel to the action execution makes sense though as that could take a lot longer than a heartbeat timeout so the server does not drop the connection meanwhile – we cannot assume that if an action execution is happening means the client is running.

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -95,6 +95,30 @@ def forward_request(base_url: str, payload: BodyPayload) -> requests.Response:
...
+    pong_queue: asyncio.Queue,
+    no_connection_queue: asyncio.Queue,
+):
+    while True:
+        await asyncio.sleep(2)
+        log.debug("️➡️ outgoing ping")
+        await ws.send("ping")
+
+        try:
+            await asyncio.wait_for(pong_queue.get(), timeout=2)
```
mkorpela, line 108: So the wait between pings is actually 2 seconds + WAIT FOR PONG

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -95,6 +95,30 @@ def forward_request(base_url: str, payload: BodyPayload) -> requests.Response:
...
 
+async def handle_ping_pong(
+    ws: websockets.WebSocketClientProtocol,
+    pong_queue: asyncio.Queue,
+    no_connection_queue: asyncio.Queue,
+):
+    while True:
+        await asyncio.sleep(2)
+        log.debug("️➡️ outgoing ping")
+        await ws.send("ping")
```
mkorpela, line 105: This might fail if the connection is broken - @ovatto points out.

action_server/src/robocorp/action_server/_server_expose.py
ktaube: It can be up to 2 extra seconds, yes

action_server/src/robocorp/action_server/_server_expose.py
ktaube: Then we will catch it outside the websocket connection context and retry.

ktaube: Also, if the connection is broken the `websocket.connect` will throw, so this won't happen.

mkorpela:
Splitting the function to smaller pieces would help readability.

fabioz:
> Yeah, it could be simplified. But currently, this works and would fix the re-connection issues. Not sure I can rework this before the release.

Ok, let's merge this is as, we can rework afterwards.

fabioz (approved):

PR #157, opened 2024-01-15
Chore:[Action Server] Improve output logging format

rihardsgravis:
Note: This PR is highly opinionated, so all changes are welcome for discussion!

- Add `rich` library to highlight and format the cli output:
- Mark some of startup messages as `debug` level since they do give less useful information to the user (e.g. logging current yaml file path or websocket connection state to the UI)
- Highlight Action Server URL's as the most important startup information
- As a user request - override Uvicorn default startup message to replace it with an Action Server message - this one is done a bit hacky by overriding a class function, but I could not find a more cleaner way to do it. Printing the URL twice (uvicorn and action server) seems a worse alternative.
- Update uvicorn request formatting to display a timestamp also
image
- Update `_run_action_in_thread` to fail with an `HTTPException` instead of the original error so the endpoint fails correctly with 500 error and action server does not show the error traceback to `_run_action_in_thread` which gives no useful information to the user:
image

### DIFF
--- a/action_server/poetry.lock
+++ b/action_server/poetry.lock
@@ -912,6 +912,21 @@ snowballstemmer = ">=2.2.0"
 toml = ["tomli (>=1.2.3)"]
 
 [[package]]
+name = "pygments"
+version = "2.17.2"
+description = "Pygments is a syntax highlighting package written in Python."
+optional = false
+python-versions = ">=3.7"
+files = [
+    {file = "pygments-2.17.2-py3-none-any.whl", hash = "sha256:b27c2826c47d0f3219f29554824c30c5e8945175d888647acd804ddd04af846c"},
+    {file = "pygments-2.17.2.tar.gz", hash = "sha256:da46cec9fd2de5be3a8a784f434e4c4ab670b4ff54d605c4c2717e9d49c4c367"},
+]
+
+[package.extras]
+plugins = ["importlib-metadata"]
+windows-terminal = ["colorama (>=0.4.6)"]
+
+[[package]]
 name = "pylint"
 version = "2.17.7"
 description = "python code static checker"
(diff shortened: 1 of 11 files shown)


ktaube (approved):
So nice! 💅💪

rihardsgravis:
@fabioz I updated linux (and on my dev machine thus also mac) tests  where stdout and stderr where incorrectly switched places [here](https://github.com/robocorp/robocorp/blob/34b8c1367b8b6ce82fa17f06f55184e3f1c862c2/action_server/src/robocorp/action_server/_robo_utils/process.py#L117), but now the Windows tests are failing because of that 🤔  I assume the tests where successful before as all the output went to stderr (which was then incorrectly streamed to stdout), but the new logger now correctly streams to stdout.
I have no access to a Windows machine and have trouble understanding from where the platform diff is coming from. Could I ask you for some debugging help on this?

fabioz:
I'm not against better looking logs, but I'm -1 on having the `rich` dep just for that... In my machine the lib+its deps bring around 10MB of dependency code, which I think is too much for the colors.

We do have in robocorp.log a utility module to print with colors already in https://github.com/robocorp/robocorp/blob/master/log/src/robocorp/log/console.py -- maybe that could be adapted for this use-case (I'm Ok in copying that and changing as needed inside the action-server code if it means not having the `rich` dep).

The other thing is that I've changed the way that the action-server process talks to subprocesses and it'll use the same approach as VSCode where the main process communicates with the subprocesses using stdin and stdout to avoid having a socket which is prone to firewall issues (and that'll have conflicts with the changes in this branch)

-- another note: we have to make sure that we don't really use the windows APIs to set colors as when we have multiple actions running in parallel that definitely won't work well.

fabioz:
> @fabioz I updated linux (and on my dev machine thus also mac) tests where stdout and stderr where incorrectly switched places [here](https://github.com/robocorp/robocorp/blob/34b8c1367b8b6ce82fa17f06f55184e3f1c862c2/action_server/src/robocorp/action_server/_robo_utils/process.py#L117), but now the Windows tests are failing because of that 🤔 I assume the tests where successful before as all the output went to stderr (which was then incorrectly streamed to stdout), but the new logger now correctly streams to stdout.
> I have no access to a Windows machine and have trouble understanding from where the platform diff is coming from. Could I ask you for some debugging help on this?

Sure, but this changed quite a bit already in the new code, so, I'd rather merge the changes I have in place for dealing with the actions being pre-loaded and only then check if this is still an issue.

rihardsgravis:
> I'm not against better looking logs, but I'm -1 on having the `rich` dep just for that... In my machine the lib+its deps bring around 10MB of dependency code, which I think is too much for the colors.
> 
> We do have in robocorp.log a utility module to print with colors already in https://github.com/robocorp/robocorp/blob/master/log/src/robocorp/log/console.py -- maybe that could be adapted for this use-case (I'm Ok in copying that and changing as needed inside the action-server code if it means not having the `rich` dep).
> 
> The other thing is that I've changed the way that the action-server process talks to subprocesses and it'll use the same approach as VSCode where the main process communicates with the subprocesses using stdin and stdout to avoid having a socket which is prone to firewall issues (and that'll have conflicts with the changes in this branch)
> 
> -- another note: we have to make sure that we don't really use the windows APIs to set colors as when we have multiple actions running in parallel that definitely won't work well.

Unfortunately the `robocorp.log` is too basic and would not allow to do simple stuff like coloring a single log message in two colors. I would argue that colors are not the only thing such library would bring to the logs, this is just the initial implementation – we can have more rich (pun intended :D) developer experience by introducing status bars, panels, tables and other interface elements.

Let me investigate on more lightweight libraries.

PR #159, opened 2024-01-16
Feat: [Action Server] Server URL param

rihardsgravis:
Add `--server-url` cli parameter that adds the URL to the OpenAPI spec `servers` section. By default, the localhost url is added:
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Robocorp Actions Server",
    "version": "0.1.0"
  },
  "servers": [
    {
      "url": "http://localhost:8080"
    }
  ]
```

The `servers` section entry is required for AI applications to acknowledge the API server endpoint URL. This allows a user to define the url when starting the server (e.g. by passing the final public url via an env environment on cloud deployments).

No special treatment is required when using `--expose` flag as the robocorp.link tunnel handles adjusting the OpenAPI schema.

### DIFF
--- a/action_server/src/robocorp/action_server/_app.py
+++ b/action_server/src/robocorp/action_server/_app.py
@@ -14,7 +14,9 @@ LOGGER = logging.getLogger(__name__)
 @cache
 def get_app():
     settings = get_settings()
-    app = FastAPI(title=settings.title)
+
+    server = {"url": settings.server_url}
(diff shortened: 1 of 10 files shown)


ktaube (approved):
LGTM! 
Sidenote: haveing these as env variables in `.env` file would also be nice. Especially the `api-key` argument.

fabioz (approved):

PR #161, opened 2024-01-16
Feat: [Action Server] Move route auth to app level

rihardsgravis:
Move action execution authorization to the application level instead of exposed server only. This enables the user passed API key `--api-key=XXX` parameter to be used by Action Server for authorization also for local requests.  Current use of `--expose` is not changed and the API key is persisted.

This feature is required to enable authorization without using `--expose`, for example, when deploying the Action Server manually.

Added also the option to input the API key in the Action Server UI.

### DIFF
--- a/action_server/src/robocorp/action_server/_robo_utils/auth.py
+++ b/action_server/src/robocorp/action_server/_robo_utils/auth.py
@@ -1,5 +1,3 @@
-import os
-from typing import Optional
 from pathlib import Path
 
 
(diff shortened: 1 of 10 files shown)


ktaube (approved):

action_server/src/robocorp/action_server/_robo_utils/auth.py
```diff
@@ -1,4 +1,27 @@
...
+
+
 def generate_api_key():
     import secrets
 
     return secrets.token_urlsafe(32)
+
+
+def get_api_key(datadir: Path) -> str:
+    api_key_path = os.path.join(datadir, ".api_key")
```
fabioz: Minor: as you're dealing with a `Path`, you could use `Path` methods.

i.e.:

`api_key_path = datadir / ".api_key"`

then to read:

`api_key_path.read_text('utf-8')`

and to write:

`api_key_path.write_text(api_key, "utf-8")`

fabioz (approved):

PR #162, opened 2024-01-16
feat: refactor expose ping logic

ktaube:
- Now uses `websockets` re-connection mechanisms
- Refactored payload handling to separate functions
- Workaround for `websockets` logger to not print stack trace when re-connecting

## Preview

https://github.com/robocorp/robocorp/assets/1479451/bebb9060-8c92-4138-a274-4eb1402c3ce6

### DIFF
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -189,7 +189,6 @@ async def expose_server(
     pong_queue: asyncio.Queue = asyncio.Queue(maxsize=1)
 
     async def listen_for_requests() -> None:
-        max_retries = 3
         retry_delay = 3
         retries = 0
 
@@ -253,7 +252,7 @@ async def expose_server(
                                 continue
             except websockets.exceptions.ConnectionClosedError:
                 retries += 1
-                log.info(f"Reconnecting to expose server ({retries}/{max_retries})")
+                log.info("Lost connection. Reconnecting to expose server...")
                 await asyncio.sleep(retry_delay * retries)
                 continue
             except socket.gaierror:
--- a/action_server/src/robocorp/action_server/_server_expose.py
+++ b/action_server/src/robocorp/action_server/_server_expose.py
@@ -15,9 +15,24 @@ from robocorp.action_server._robo_utils.process import exit_when_pid_exists
 
 log = logging.getLogger(__name__)
 
+
 # disable websockets logger. It's too verbose
+class NoExceptionFormatter(logging.Formatter):
+    def format(self, record):
+        record.exc_info = None
+        return super().format(record)
+
+
+console_handler = logging.StreamHandler()
+console_handler.setLevel(logging.DEBUG)
+
+formatter = NoExceptionFormatter("%(message)s")
+console_handler.setFormatter(formatter)
+
+
 websockets_logger = logging.getLogger("websockets")
-websockets_logger.setLevel(logging.ERROR)
+websockets_logger.addHandler(console_handler)
+websockets_logger.propagate = False
 
 
 class SessionPayload(BaseModel):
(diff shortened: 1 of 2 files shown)


action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -124,180 +180,92 @@ async def expose_server(
...
+        headers = (
+            {
+                "x-session-id": session_payload.sessionId,
+                "x-session-secret": session_payload.sessionSecret,
+            }
+            if session_payload
+            else {}
+        )
+
+        async for ws in websockets.connect(
```
ktaube, line 177: Uses websockets [own mechanisms](https://websockets.readthedocs.io/en/stable/faq/client.html#how-do-i-reconnect-when-the-connection-drops) to reconnect.

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -98,23 +118,74 @@ def forward_request(base_url: str, payload: BodyPayload) -> requests.Response:
...
 
-class ExposePingTimeoutError(Exception):
-    pass
+def handle_session_payload(
+    session_payload: SessionPayload, api_key: str | None, expose_url: str, datadir: str
+):
+    url = f"https://{session_payload.sessionId}.{expose_url}"
+    log.info(f"🌍 URL: {url}")
+    if api_key is not None:
+        log.info(
```
fabioz: Question: shouldn't the api_key be checked here?

action_server/src/robocorp/action_server/_server_expose.py
```diff
@@ -15,6 +16,25 @@
...
+
+console_handler = logging.StreamHandler()
+console_handler.setLevel(logging.DEBUG)
+
+formatter = NoExceptionFormatter("%(message)s")
+console_handler.setFormatter(formatter)
+
+
+websockets_logger = logging.getLogger("websockets")
+websockets_logger.addHandler(console_handler)
```
fabioz: I think this shouldn't be done here... the logging is configured at `robocorp.action_server.cli._setup_stdout_logging` and `robocorp.action_server.cli._setup_logging`, so, this probably belongs there...

action_server/src/robocorp/action_server/_server_expose.py
ktaube: Not relevant after https://github.com/robocorp/robocorp/pull/161

action_server/src/robocorp/action_server/_server_expose.py
ktaube: `websockets` has its own re-connect logic, but it prints the failure logs with the stack trace making the output too verbose. This is just a workaround to stop it from printing the stack trace. See [websockets code](https://github.com/python-websockets/websockets/blob/35bc7dd8288445289134c335aae8af859862ccd1/src/websockets/legacy/client.py#L606) 

So I don't think this workaround should be moved to `robocorp.action_server.cli`.

fabioz: The issue here is that the code is adding directly a logger and stopping propagation and the logger is redirected to stdout.

The action server actually adds 2 handlers, one for the stdout and another to a rotating file handler. The code here just adds one related to the stdout and stops the propagation for the root in the stdout, but then the rotating file handler doesn't get any notification.

This means that logs will always be shown in the console the way it's done here and not in the file and the log level here doesn't respect the log level that was configured in `_setup_stdout_logging`/`_setup_logging`, which is not what's expected (it's expected that the log level will respect the user setting and also it should redirect to both, stdout and the file).

ktaube: Okay, any pointers how to

1. keep `websockets` re-connect logging
2. remove the stack trace

without my changes?

fabioz: Maybe you could change the current formatters to format differently (i.e.: remove the exception info) based on the `module` and `funcName` available?  (in `_setup_stdout_logging`/`_setup_logging`)

i.e.: Create a new formatter which wraps an existing formatter and just implements `format` and removes the exception info if the record module/funcName match something you expect. That way you don't need to add a new handler, just change the existing formatters to be that formatter-wrapper.

Does it make sense?

ktaube:
Had a call with Fabio and now I understand what changes I need to do for logging. Will do in separate PR.

fabioz (approved):

PR #164, opened 2024-01-17
Chore:[Action Server] Improve output logging format

rihardsgravis:
A more lightweight approach by using [termcolor](https://pypi.org/project/termcolor/) library only to colorize and style terminal output:
- The colors are disabled when the log level is `DEBUG`. The [NO_COLOR](https://no-color.org) env variable is also respected
- The ANSI codes are striped from the `RotatingFileHandler` file logs
- Mark some of startup messages as `debug` level since they do give less useful information to the user (e.g. logging current yaml file path or websocket connection state to the UI)
- Highlight Action Server URLs as the most important startup information
- Update uvicorn request formatting to display a timestamp also

image

- Update `_run_action_in_thread` to fail with an `HTTPException` instead of the original error so the endpoint fails correctly with 500 error and action server does not show the error traceback to `_run_action_in_thread` which gives no useful information to the user:
image

### DIFF
--- a/action_server/src/robocorp/action_server/_server.py
+++ b/action_server/src/robocorp/action_server/_server.py
@@ -245,15 +245,17 @@ def start_server(
     def _on_started_message(self, **kwargs):
         (host, port) = _get_currrent_host()
 
+        padding_bottom = "" if (api_key or expose) else "\n"
         log.info(
             colored("\n  ⚡️ Action Server started at: ", "green", attrs=["bold"])
-            + colored(f"http://{settings.address}:{port}", "light_blue")
+            + colored(f"http://{settings.address}:{port}{padding_bottom}", "light_blue")
         )
 
         if api_key:
+            padding_bottom = "" if expose else "\n"
             log.info(
                 colored("  🔑 API Authorization key: ", attrs=["bold"])
-                + f'{{ "Authorization": "Bearer {api_key}"}}'
+                + f'{{ "Authorization": "Bearer {api_key}"}}{padding_bottom}'
             )
 
     async def _on_startup():
--- a/action_server/src/robocorp/action_server/_server.py
+++ b/action_server/src/robocorp/action_server/_server.py
@@ -239,24 +239,30 @@ def start_server(
             settings.expose_url,
             settings.datadir,
             str(expose_session),
+            api_key,
         ]
         expose_subprocess = subprocess.Popen(args)
 
     def _on_started_message(self, **kwargs):
         (host, port) = _get_currrent_host()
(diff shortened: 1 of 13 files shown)


action_server/src/robocorp/action_server/_settings.py
```diff
@@ -153,3 +154,40 @@ def get_settings() -> Settings:
     if _global_settings is None:
         raise AssertionError("It seems that the settings have not been setup yet.")
     return _global_settings
+
+
+UVICORN_LOG_CONFIG = {
```
fabioz: I think all all log configurations should be done in `robocorp.action_server.cli._setup_stdout_logging` and `robocorp.action_server.cli._setup_logging` (having log configurations scattered in different places makes it harder to reason about it).

i.e.: I'm not sure what's the final result combining the loggers we setup there with the loggers setup here, especially considering the additional handlers added here (stream handlers are added, but what about the handler that redirects to the file, why would `uvicorn.access` no longer be propagated to the file -- is this something we want?)

action_server/src/robocorp/action_server/_server.py
```diff
@@ -243,11 +244,16 @@ def expose_later(loop):
...
+
+        log.info(
+            colored("\n  ⚡️ Action Server started at: ", "green", attrs=["bold"])
+            + colored(f"http://{settings.address}:{port}", "light_blue")
+        )
 
         if api_key:
             log.info(
-                f'  🔑 API Authorization key: {{ "Authorization": "Bearer {api_key}"}}'
+                colored("  🔑 API Authorization key: ", attrs=["bold"])
```
kariharju: Question on: 
`API Authorization key: { "Authorization": "Bearer Kt06mKrycBAYWtZ8QwiM_EfBL4O9CWbSKWuTkow0Z98"}`
..where do we need this in the raw format?

If most of the GPT config are implemented in a way that user selects auth. type from a drop-box and copies in "API Key", then should we parse and present these in that from?

rihardsgravis: Valid point, updated it to be just:
image

action_server/src/robocorp/action_server/_settings.py
rihardsgravis: Refactored handling Uvicorn logs – instead of defining Log formatting as suggested by Uvicorn, added a custom Formatter that handles Uvicorn access logs explicitly and a logging filter that does filter out all other Uvicorn messages – these are there if logging level is DEBUG.

fabioz (approved):

PR #181, opened 2024-01-23
Filter endpoints from Action Server OpenAPI spec

rihardsgravis:
Remove all endpoints from the OpenAPI spec that are not required for the AI use. The endpoints are still there and can be used "when you know what you are doing", but AI applications should not be exposed to these.

- Remove `/runs` and `/api/action_packages` from OpenAPI spec
- Removed `base_log.html` endpoint at all -  @fabioz I could not find a reference where it is used – maybe there is some undocumented use for it?


fabioz:
@rihardsgravis the `base_log.html` is really not used right now... The idea would be getting the log contents and updating it in real-time based on the raw log contents for the run, but that's really not done right now.

How would you feel about making this a toggle? (by default only the part related to AI would be shown and based on a command line argument -- say `--full-open-api-spec` -- it'd show the full spec, which could be useful for clients that want to automate that)

rihardsgravis:
> How would you feel about making this a toggle? (by default only the part related to AI would be shown and based on a command line argument -- say `--full-open-api-spec` -- it'd show the full spec, which could be useful for clients that want to automate that)

@fabioz refactored the OpenAPI spec endpoints to be excluded/included based on the `--full-openapi-spec` flag 👍

PR #182, opened 2024-01-23
Action Server frontend dependencies update

rihardsgravis:
Action Server frontend dependencies update:
- Major ` vite-plugin-singlefile` version update does not require breaking changes as we already use vite v.5

Closes #179


fabioz (approved):
LGTM!

PR #188, opened 2024-01-25
Improve issue templates

### LINKED ISSUE #170: Improve the "submit issue" template
### Description

It's a bit annoying that the `target library` is added as a list in the description instead of a label in the issue.

My suggestion is using something as: https://github.com/marketplace/actions/advanced-issue-labeler so it'd be a combo and it'd automatically label it accordingly.

Also, the same should be done for feature/bug: it should be a label, not appear in the title.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

cmin764:
Closes #170 

Drops the following in favor to labels:
- Tags from the title
- Checkboxes list of libs -> replaced by dropdown

~Can be further [improved](https://github.com/robocorp/robocorp/issues/170#issuecomment-1910030656) with additional automation by someone willing to take the effort. (IMHO I think it's too early to do that now)~

Selecting an affected package from the drop-down will automatically label the issue appropriately.


fabioz:
One thing is that this doesn't nudge toward selecting the initial labels (the previous way was annoying because it created tasks which didn't fit in, but having the "nudge" to select initially was nice).

I think https://github.com/marketplace/actions/advanced-issue-labeler could fit the bill to do both...

.github/ISSUE_TEMPLATE/bug-report.yaml
```diff
@@ -1,27 +1,35 @@
 name: 🐛 Bug report
 description: Submit a bug report to help us make Robocorp better
-title: "[Issue]: "
+labels: [ bug ]
```
cmin764, line 3: If this really works, then such label should be automatically applied when a new issue is created with such template.

rihardsgravis (approved):
👏

.github/ISSUE_TEMPLATE/bug-report.yaml
```diff
@@ -1,27 +1,35 @@
...
 body:
-  - type: checkboxes
-    id: library
+  - type: dropdown
+    id: components
```
cmin764, line 6: Libraries aren't the only things we can improve in this repo, there's also the Action Server alone, integration tests, chores or dev utils. (and what knows more)

.github/ISSUE_TEMPLATE/bug-report.yaml
```diff
@@ -1,27 +1,35 @@
...
+        - integration_tests
+        - log
+        - log_pytest
+        - meta
+        - storage
+        - tasks
+        - templates
+        - vault
+        - windows
+        - workitems
```
cmin764, line 29: Much easier to just multiple select the affected packages (or not) as project components. Then is the job of the issue labeler to apply the right labels given these.

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -1,27 +1,35 @@
...
-          required: false
-        - label: robocorp-tasks
-          required: false
-        - label: robocorp-vault
-          required: false
-        - label: robocorp-windows
-          required: false
-        - label: robocorp-workitems
-          required: false
+        - Other
```
cmin764, line 12: There can be a case where none of the selections applies.

.github/workflows/advanced-issue-labeler.yml
cmin764, line 1: This is the policy of the issue labeler flow. It looks at the values selected from the `components` dropdown in the issue template, and then maps those with the label names described here through the matching keys.

E.g.: **components**: `action_server`, `actions` -> Labels: _Action Server_, _Library_, _robocorp-actions_

.github/workflows/issue_labeler.yml
cmin764, line 1: Adds two steps with **[advanced-issue-labeler](https://github.com/marketplace/actions/advanced-issue-labeler)** for both the bug-report and the feature-request templates.

cmin764 (commented):
Left a self-review to explain the whole flow.

Will merge this as it is in order to test the flow and if something doesn't work as expected, we can come back to fix this when this will become a priority again.

.github/workflows/issue_labeler.yml
```diff
@@ -0,0 +1,49 @@
...
+          template-path: .github/ISSUE_TEMPLATE/bug-report.yml
+
+      - name: Set bug labels based on the component field
+        uses: redhat-plumbers-in-action/advanced-issue-labeler@v3.0.0
+        with:
+          issue-form: ${{ steps.bug-issue-parser.outputs.jsonString }}
+          section: components
+          block-list: |
+            Other
+          token: ${{ secrets.CI_GITHUB_TOKEN }}
```
cmin764, line 34: @fabioz / @rihardsgravis hopefully this is the right token (org level) to be used with such flows.

PR #206, opened 2024-01-30
Add cryptography to 3rd party library docs

### LINKED ISSUE #216: 3rd-party: cryptography

OvidiuCode:


docs/3rd_party/cryptography/README.md
```diff
@@ -0,0 +1,24 @@
+# Cryptography
+
+The `cryptography` library is a robust toolkit for implementing secure encryption, decryption, and cryptographic protocols in applications.
+
+## Usage
+
+```python
+from cryptography.hazmat.primitives.asymmetric import rsa
```
kariharju, line 8: @OvidiuCode & @cmin764 good to check in this example what is the latests recommended encryption algorithm, key sizes etc. so that we don't guide anything obsolete.

cmin764: According to some SSH best practices (not sure if relevant), an RSA of 2048 or even 4096 bit is still considered secure.

Although there are these more exotic eliptic curves ones, like ECDSA which are stronger at brute-force. ([context](https://security.stackexchange.com/a/144044))
> Generating Ed25519 keys is faster compared to traditional RSA keys.

[It’s 2023. You Should Be Using an Ed25519 SSH Key (And Other Current Best Practices)](https://www.brandonchecketts.com/archives/its-2023-you-should-be-using-an-ed25519-ssh-key-and-other-current-best-practices)

So this seems to be the new kid on the block.

cmin764 (approved):
Wouldn't hurt to see some more cases, like:

1. Symmetric encryption (e.g. AES)
2. Hashing a password (e.g. SHA-256) -- which you used already, but not in this context
3. Signing something (encrypting a hash with the **private** (not public) key)

But as always, these are nice to have, no immediate real customer impact with such.

PR #218, opened 2024-01-31
Dependency explicit requirement notice for availability in docs

### LINKED ISSUE #198: Let users know about the required dependencies in the docs
Example for taking into use `robocorp.windows`:
> In order to use this, you have to make it available in your Python environment by listing `robocorp-windows` as a requirement in your dependencies configuration file.
> - _conda.yaml_ for an automation Tasks Package
> - _requirements.txt_, _pyproject.toml_ etc. for the rest

Do the same with `browser` and `excel` as well.

cmin764:
Closes #198

Adds notices about the need to include `robocorp-*` packages which aren't included by default in `robocorp` in case you want to use such libraries.

----

Not bad for ReMark:

Screenshot 2024-01-31 at 5 41 44 PM


windows/src/robocorp/windows/_vendored/uiautomation/uiautomation.py
cmin764, line 1: @fabioz Since this is vendorized code, shall we put some sort of rule for Ruff to not touch the sources in this directory? (although `check-all` command was complaining about it)

fabioz: Yes.

MonicaBarna (approved):

PR #226, opened 2024-02-06
Store the exception on work item in case it is marked as failed

### LINKED ISSUE #169: Remember in the `Input` work item the error on a failed release
### Description

It will be very convenient to know more about a failed work item without consulting Control Room for it, that being the exception that was created for it on a `.fail()` type of release. This information is sent as meta-data to CR when the input work item gets released as failed.

### Describe the solution you'd like

We can save this into an `_exception` private attribute and expose it through a property with the same name. (similar to `_state`)

- File to change: `workitems/src/robocorp/workitems/_workitem.py` -> `fail()` method

### Describe alternatives you've considered

An alternative would be for the method to return (instead of setting) such a generated error, but that means you have to remember such error for every released failed item and might make the API of this release method a little bit unaligned given the `.done()` success release (and the rest of the flow).

OvidiuCode:
Closes #169 

Add exception property to work items.

### DIFF
--- a/workitems/docs/CHANGELOG.md
+++ b/workitems/docs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.4.3 - 2024-02-06
+## 1.4.3 - 2024-02-07
 
 - Add exception property to workitems
 
--- a/workitems/docs/CHANGELOG.md
+++ b/workitems/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.4.3 - 2024-02-06
+
+- Add exception property to workitems
+
 ## 1.4.2 - 2024-01-14
 
 - Fix main README and update docs.
--- a/workitems/tests/workitems_tests/test_workitems.py
+++ b/workitems/tests/workitems_tests/test_workitems.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+from collections import OrderedDict
 from contextlib import contextmanager
 from pathlib import Path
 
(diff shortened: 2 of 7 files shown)


workitems/tests/workitems_tests/test_workitems.py
```diff
@@ -128,16 +130,22 @@ def test_input_pass(inputs, adapter):
...
     assert len(adapter.releases) == 0
 
-    item.fail(ExceptionType.BUSINESS, "MY_ERR_CODE", "Something went oopsy")
+    exception = {
+        "type": ExceptionType.BUSINESS,
+        "code": "MY_ERR_CODE",
+        "message": "Something went oopsy",
+    }
+
+    item.fail(*exception.values())
```
cmin764, line 141: Careful with this one, as there's no guarantee that the `.values()` are returned in order compared to how the `exception` dictionary was declared. I think [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict) may save you here.

cmin764 (changes requested):
LGTM!

Requested change for the `OrderedDict` thing to diminish the risk of a flaky test.

----

Some PR hygiene:

pr-hygiene

workitems/docs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.4.3 - 2024-02-06
+
+- Add exception property to workitems
+
```
cmin764: This date is good, but if you plan to make the release tomorrow, you should update it:
```suggestion
## 1.4.3 - 2024-02-07

- Support for retrieving the error through the `exception` property of failed input work items.

```

But if you release today, all good then.

cmin764 (approved):
LGTM! And if you make the release right afterwards, you're good to go.

workitems/docs/CHANGELOG.md
OvidiuCode: I'll make it tomorrow. Will update

Issue #92, opened 2023-11-09
Improve the Developer Experience


cmin764:
A couple of improvements we can bring to `invoke <command>` for a better DX:
- [x] Ensure that initial requirements (Invoke & Poetry) can be brought without creating a Poetry _.venv_ first.
- [x] #255
- [x] Commands depending on the local virtual environment should be going through Poetry's context. (e.g. `inv make-release` fails to correctly detect the current version, where `poetry run inv make-release` works)
- [x] #262
- [x] #264
- [ ] #261
- [ ] #263 
- [x] #287 
- [x] #299
- [x] #300
- [x] #337 
- [ ] #267

📝 At the end, we should have a clear development guide teaching you in what order these commands should be run during the development journey.

fabioz:
Humm, I just don't use poetry at all after the install with conda.

My workflow is something as:

```
conda create -n robocorp-log python=3.10 -c conda-forge
conda activate robocorp-log
pip install poetry
poetry install
```

And then completely forget about poetry, the only thing being that I need to remember to activate the env and maybe call poetry install again later on (so, the proposed improvements would actually break the conda usage).

cmin764:
> Humm, I just don't use poetry at all after the install with conda.
> 
> My workflow is something as:
> 
> ```
> conda create -n robocorp-log python=3.10 -c conda-forge
> conda activate robocorp-log
> pip install poetry
> poetry install
> ```
> 
> And then completely forget about poetry, the only thing being that I need to remember to activate the env and maybe call poetry install again later on (so, the proposed improvements would actually break the conda usage).

Now that makes sense. Still, do you mind if we try to support different workflows without breaking yours?

I find it a little bit easier to just:
1. `cd` into the package you want to develop.
2. `inv install`: Poetry updates & installs the package in dev mode in the local _.venv_ folder.

So there's no need to remember to create & activate virtual environments, everything being tackled through Invoke & Poetry by just looking at the project folder you're currently in.

To achieve this, we need to make sure that:
1. All Invoke commands will run through the Poetry context. (so Invoke & Poetry have to be installed **once** first -- can be added as requirements in the _devutils_ project)
2. Main interpreter to be used to create the venv by Poetry can be isolated initially at repo root level (not project/package) with a tool like Pyenv. (see [details](https://github.com/robocorp/rpaframework/blob/master/docs/source/contributing/development.md#using-pyenv-on-any-platform-to-manage-interpreters))

----

[...]

PR #232, opened 2024-02-07
Fix workitems workflow and issue labeling policy

cmin764:
- Moves the Issue labeling policy in its right path. (so we won't run it for other actions)
- Fixes the release Poetry invocation for `workitems`.


OvidiuCode (approved):

cmin764:
Will merge these as it is right now, then we come back and fix the source of the issue later.

Related to: #92

PR #233, opened 2024-02-07
Improve the developer experience through better Invoke commands

### LINKED ISSUE #92: Improve the Developer Experience

### LINKED ISSUE #255: Improve Invoke commands
According to [Notion](https://www.notion.so/robocorptechnologies/DX-improvements-519c97b7bf7c4bf097b2ca0885a4e51b?pvs=4)

cmin764:
Closes #255 

- More robust way of defining the initial dependencies which will setup the package environment. (plain `inv <cmd>` working in all the packages)
- Improved Invoke commands according to [Notion](https://www.notion.so/robocorptechnologies/DX-improvements-519c97b7bf7c4bf097b2ca0885a4e51b?pvs=4).
- Fixes the `meta` package way of importing and using `devutils`.


devutils/pyproject.toml
```diff
@@ -19,9 +20,7 @@ pytest-regressions = "1.0.6"
 pydocstyle = "^6.3.0"
 isort = { version = "^5.12.0", python = "^3.8" }
 types-invoke = "^2.0"
-invoke = "^2.0"
 lazydocs = { git = "https://github.com/robocorp/lazydocs.git" }
```
fabioz, line 24: I'm not sure I follow, what do we gain by removing things from pyproject.toml?

excel/tasks.py
```diff
@@ -1,14 +1,23 @@
...
 try:
-    import devutils
+    from devutils import invoke_utils
 except ImportError:
-    devutils_src = Path(__file__).absolute().parent.parent / "devutils" / "src"
-    assert devutils_src.exists(), f"{devutils_src} does not exist."
-    sys.path.append(str(devutils_src))
-
-from devutils.invoke_utils import build_common_tasks
+    invoke_utils = SourceFileLoader(
```
fabioz: I'm not sure this is an improvement... it works because the `invoke_utils` doesn't import anything outside of that file at this point, but it'll break when that happens.

devutils/pyproject.toml
cmin764: Was thinking that this might take away from the confusion of who is responsible for providing you with the Invoke tool. Is it yourself initially or the Poetry-based venv after a `poetry install`? (this way you'd be sure that you should get Invoke from the _requirements.txt_ only and that is something living outside your package _.venv_)

Also it helps with pinning & maintaining a compatible version in a single place, which is _requirements.txt_. But if you'll need to use Invoke from Poetry's context (e.g.: `inv <task>` -> `poetry(..., "inv", ...)`), then yes, it might make sense to keep that dep here if the initial executable isn't reachable anymore.

Will add it back if it doesn't produce any harm or confusion, as I don't have a strong opinion either. And now this stems another question:
> What do we gain by adding unnecessary dependencies in **devtools**' _pyproject.toml_?

excel/tasks.py
cmin764: Correct, I missed that!

Just that changing the `sys.path` at runtime (and such change has to be maintained in a dozen of different _tasks.py_ files) didn't look like a better idea than this newer suggestion I added. (and improved post-review)

----

I added a few more fixes to strengthen the solution and make it cleaner, so that:
- Any _tasks.py_ file doesn't even need to know about the files/modules we write in _devutils_, it is sufficient now to refer to: `"devutils" / "src" / "devutils" / "__init__.py"`
- Any "publicly" exposed function at package init level is directly usable by any importer: `devutils.build_common_tasks()`
- The _invoke_utils.py_ file is the only one in charge now for managing `sys.path` in a single place, with a path relative to itself, and now we can even do imports like: `from devutils.rcc import build_subprocess_kwargs`

What do you think now, isn't [this](https://github.com/robocorp/robocorp/blob/chore/dx-improvements/excel/tasks.py#L20-L23) more cleaner, robust and easier to maintain now in all those _tasks.py_ files we have in every package?

devutils/src/devutils/invoke_utils.py
```diff
@@ -12,9 +12,15 @@
 
 from invoke import task
 
-ROOT = Path(__file__).parent.parent.parent
+
+SRC = Path(__file__).resolve().parent.parent
+ROOT = SRC.parent
 REPOSITORY_URL = "https://github.com/robocorp/robocorp/tree/master/"
 
+sys.path.append(str(SRC))
```
bogdancondurache: one idea, might be premature optimisation at this point, would be to have this part in the __init__.py file and then you would be able to use `devutils` imports from any file, not just this one. Of course, at the moment it's just this file, but in the future if somebody adds a new file and the imports won't work he/she might be very confused as to why :)

devutils/pyproject.toml
fabioz: > What do we gain by adding unnecessary dependencies in devtools' pyproject.toml?

Nothing, but as you mentioned before:

> But if you'll need to use Invoke from Poetry's context (e.g.: inv <task> -> poetry(..., "inv", ...)), then yes, it might make sense to keep that dep here if the initial executable isn't reachable anymore.

So, we do get something here (although our automations using `inv` make it so that it doesn't need the poetry env active, sometimes I do expect `inv` or `poetry` to be in the env if I'm doing things that are not initially in our automations).

excel/tasks.py
```diff
@@ -1,14 +1,21 @@
-import sys
+from importlib.machinery import SourceFileLoader
 from pathlib import Path
 
-# Add the devutils even if the poetry env isn't setup (to do a 'inv devinstall').
+# Since the `robocorp-devutils` dependency is available in the local Poetry isolated
+#  environment only, we won't have it available with Invoke commands running outside
+#  such environment.
```
bogdancondurache:
```suggestion
# environment only, we won't have it available with Invoke commands running outside
# such environment.
```

devutils/src/devutils/invoke_utils.py
cmin764: Yes and thank you for the notice!

I'll revert all of these anyway as it was proved that they bring more complexities than clarity. Will ask for a proper review once I mark this as non-draft anymore :).

excel/tasks.py
cmin764: There's a rule in PyCharm where such indentation separates a contiguous comment block from the other comments.

Highlight can be observed with TODO and FIXME:
Screenshot 2024-02-08 at 4 05 50 PM
Screenshot 2024-02-08 at 4 05 57 PM

excel/tasks.py
cmin764: Reverted as it was proved that `devutils` isn't expected to be a cohesive Python package and my change didn't add value.

excel/tasks.py
bogdancondurache: interesting, had no idea this is a normal format

devutils/src/devutils/invoke_utils.py
```diff
@@ -12,7 +12,7 @@
 
 from invoke import task
 
-ROOT = Path(__file__).parent.parent.parent
+ROOT = Path(__file__).resolve().parent.parent.parent
```
fabioz: Please don't `resolve()` here as it'll resolve symlinks which is not what we want.

Use `absolute()` instead.

meta/tasks.py
```diff
@@ -1,18 +1,27 @@
+import sys
 from pathlib import Path
 
-from devutils.invoke_utils import build_common_tasks
 from invoke import task
 
-CURDIR = Path(__file__).absolute().parent
+ROOT = Path(__file__).resolve().parent
```
fabioz: Don't `resolve()`. Use `absolute()` instead.

Note: you almost never want to `resolve()` unless you need a unique reference to some file and want to create some sort of cache.

The reason for this is because when symlinks are resolved it'll change the view of the working dir.

For instance, if you are in a subst folder such as:

`x:`

which points to `d:\my_x_subst`

And the .pyc was generated with the file pointing to `x:\my_file.py`, if you resolve() it'll actually become `d:\my_y_subst\my_file.py` and you'd break any view which was tailored by the user for the working dir being used (at best -- probably the case here -- you'd just start getting the .pycs from a different `__file__`, but at worst you would end up having imports that no longer resolve because that different view of the structure is no longer true anymore).

devutils/src/devutils/invoke_utils.py
```diff
@@ -197,15 +200,31 @@ def _make_conda_env_if_needed():
...
-    def install(ctx):
-        """Install dependencies"""
-        _make_conda_env_if_needed()
-        poetry(ctx, "install")
+    def install(ctx, local: str = None, update: bool = False, verbose: bool = False):
+        """
+        Installs or updates dependencies.
 
-    @task
-    def devinstall(ctx):
```
fabioz, line 206: Where did `devinstall` go? It should still be there...

OvidiuCode: We delete it in favor of --local argument on install. But if you think we need a standalone command we can put it back.

fabioz: I'd rather have a standalone command -- and then you won't need to update all the github actions to use the new version too ;)

OvidiuCode: ok, I'll add it back

devutils/src/devutils/invoke_utils.py
```diff
@@ -171,12 +172,15 @@ def run(ctx, *cmd, **options):
...
 
-    def poetry(ctx, *cmd):
+    def poetry(ctx, *cmd, verbose: bool = False):
         prefix = []
         if _use_conda():
             prefix.extend(["conda", "run", "--no-capture-output", "-n", CONDA_ENV_NAME])
         prefix.append("poetry")
 
+        if verbose:
+            prefix.append("--verbose")
```
cmin764:
```suggestion
            prefix.append("-vv")
```
as one level of verbosity is not enough here. (doesn't spit too much)

devutils/src/devutils/invoke_utils.py
```diff
@@ -197,15 +201,45 @@ def _make_conda_env_if_needed():
                 )
 
     @task
-    def install(ctx):
-        """Install dependencies"""
+    def install(ctx, local: str = None, update: bool = False, verbose: bool = False):
```
cmin764:
```suggestion
    def install(ctx, local: Optional[str] = None, update: bool = False, verbose: bool = False):
```

devutils/src/devutils/invoke_utils.py
cmin764: Good point!

devutils/src/devutils/invoke_utils.py
```diff
@@ -197,15 +201,45 @@ def _make_conda_env_if_needed():
...
+            verbose: Whether to run in verbose mode.
+        """
         _make_conda_env_if_needed()
+
+        with mark_as_develop_mode(all=True):
+            poetry(ctx, "lock --no-update")
+            poetry(ctx, "install", verbose=verbose)
+
+    @contextmanager
+    def mark_as_develop_mode(projects: Optional[list[str]] = None, all=False):
```
cmin764:
```suggestion
    def mark_as_develop_mode(projects: Optional[list[str]] = None, all_packages: bool = False):
```
if we still want to stay consistent with typing. Also is not a good idea to shadow the builtin `all`.

devutils/src/devutils/invoke_utils.py
```diff
@@ -271,11 +306,16 @@ def pretty(ctx):
         poetry(ctx, f"run isort {targets}")
 
     @task
-    def test(ctx):
+    def test(ctx, test=None):
```
cmin764:
```suggestion
    def test(ctx, test: Optional[str] = None):
```

devutils/src/devutils/invoke_utils.py
```diff
@@ -390,30 +445,63 @@ def make_release(ctx):
...
+            new_content = (
+                content[:changelog_start]
+                + double_newline
+                + new_version_section
+                + content[last_version_pos:]
+            )
+
+            stream.seek(0)
+            stream.write(new_content)
+            print(print("Changed: ", file))
```
cmin764: I would also print the `unreleased_changes` as well, so I would have a confirmation of what I'm releasing next, as the file that gets changed is always the same and doesn't say too much.

excel/tasks.py
cmin764, line 1: If we like this refactor (which adds a little bit more clarity imho), then we can copy-paste it to the other _tasks.py_ files as well. (doesn't change behaviour, just structure)

devutils/src/devutils/invoke_utils.py
```diff
@@ -197,15 +201,45 @@ def _make_conda_env_if_needed():
...
+    def devinstall(ctx, verbose: bool = False):
+        """
+        Install the package in develop mode and its dependencies.
+
+        Args:
+            verbose: Whether to run in verbose mode.
+        """
         _make_conda_env_if_needed()
+
+        with mark_as_develop_mode(all=True):
```
cmin764: This `all` is a reserved Python keyword (the builtin function), I would recommend replacing it with something like `all_packages` (or similar) to avoid name shadowing.

devutils/src/devutils/invoke_utils.py
```diff
@@ -217,21 +251,22 @@ def devinstall(ctx):
...
-                            name = key[len("robocorp-") :]
-                            value = dict(path=f"../{name}/", develop=True)
-                            dependencies[key] = value
-
-            poetry(ctx, "lock --no-update")
-            poetry(ctx, "install")
+                        # robocorp-log = {path = "../log/", develop = true
+                        name = key[len("robocorp-") :]
+                        if all or name in projects:
+                            dependencies[key] = dict(path=f"../{name}/", develop=True)
```
cmin764: 🐛 Bug spotted:
```suggestion
                        if all or key in projects:
                            dependencies[key] = dict(path=f"../{name}/", develop=True)
```

- `name`: `workitems`
- `key`: `robocorp-workitems`
- `projects`: [`robocorp-workitems`]

So you should check the `key` belonging to the list, not the `name`, which always fails to be found.

cmin764 (commented):
LGTM overall and thank you @OvidiuCode for your hard work!

⚠️ Still, I want to request some changes before merging as pointed in the comments below this one.

----

Once this gets merged (or about to be merged), please check what workflows need to be changed/updated to match the new state of the upcoming Invoke commands. Something like:

```
    - name: Install poetry
       run: pipx install poetry
    ....
    - name: Install invoke/tomlkit
       run: pip install invoke tomlkit
```
->
```
    - name: Install development dependencies
       run: pipx install -Ur requirements.txt  # from devutils
```

`python -m invoke <cmd>` -> `inv <cmd>`
if works and simpler to use

```
    - name: Test
      env:
        GITHUB_ACTIONS_MATRIX_NAME: ${{ matrix.name }}
        CI_CREDENTIALS: ${{ secrets.CI_CREDENTIALS }}
        CI_ENDPOINT: ${{ secrets.CI_ENDPOINT }}
      run: python -m invoke test
    - name: '`inv lint`/`inv typecheck`, potentially fixed with `inv pretty`'
      if: always()
      run: '

        python -m invoke lint

        python -m invoke typecheck

        '
```
->
```
    - name: Lint and test
      env:
        GITHUB_ACTIONS_MATRIX_NAME: ${{ matrix.name }}
        CI_CREDENTIALS: ${{ secrets.CI_CREDENTIALS }}
        CI_ENDPOINT: ${{ secrets.CI_ENDPOINT }}
      run: inv check-all
```
since we're checking in one go now: linting, typing, tests and docs. (and we were checking linting & typing anyways for the entire matrix before)


```
    - name: Check tag version
      run: poetry run invoke check-tag-version
```
->
```
    - name: Check tag version
      run: inv check-tag-version
```

```
    - name: Build with poetry
      run: poetry build

    - name: Upload to PyPI
      run: |
        poetry config pypi-token.pypi  ${{ secrets.PYPI_TOKEN_ROBOCORP_WORKITEMS }}
        poetry publish
```
->
```
    - name: Build package and upload to PyPI
      run: |
        inv build
        inv publish --token ${{ secrets.PYPI_TOKEN_ROBOCORP_WORKITEMS }}
```
once we add such additional `pubish` Invoke command.

devutils/src/devutils/invoke_utils.py
fabioz: `print(print(...))` seems odd here...

devutils/src/devutils/invoke_utils.py
```diff
@@ -390,30 +445,63 @@ def make_release(ctx):
...
+
+        with open(file, "r+") as stream:
+            content = stream.read()
+
+            # Prepend a new section with the new version and changes
+            new_version_section = (
+                f"## {version} - {datetime.today().strftime('%Y-%m-%d')}\n\n"
+            )
+
+            changelog_start = re.search(r"# Changelog", content).end()
```
fabioz, line 497: What's the expected format of the changelog that this will act upon?

The format I was usually working on so far was:

```
# Changelog

## Unreleased

- xxx
- yyy

## 0.0.20 - 2024-01-31

- zzz
```

But then I'd expect this code here to just change the 'Unreleased' to the version here, but this isn't what the code is doing...

devutils/src/devutils/invoke_utils.py
cmin764: Yes, it prints an additional `None` because of that, just spotted it during testing.

devutils/src/devutils/invoke_utils.py
fabioz: I think it may be nicer to change the code to deal with that structure and raise an error if `## Unreleased` is not found right after `## Changelog` complaining that the changelog is not up to date.

devutils/src/devutils/invoke_utils.py
```diff
@@ -390,30 +445,63 @@ def make_release(ctx):
...
+                + new_version_section
+                + content[last_version_pos:]
+            )
+
+            stream.seek(0)
+            stream.write(new_content)
+            print(print("Changed: ", file))
+
     @task
     def set_version(ctx, version):
```
cmin764, line 515:
```suggestion
    def set_version(ctx, version: str):
```

Would be awesome if you can parse the version and check if it's numeric, as I'm always accidentally running: `inv set-version -h` to see some usage and guess what, I end up with this version:
```
__version__ = "-h"
```

You can use this [pattern](https://github.com/robocorp/rpaframework/blob/master/packages/core/src/RPA/core/certificates.py#L45): `packaging.version.parse()` to parse and check the atoms if are integers. (or at least 3 in number, as we're following semver: `major.minor.hotfix`)

excel/tasks.py
fabioz: I agree that having a variable to don't duplicate the `Path(__file__).absolute().parent` is a bit better -- but I'm also not bothered enough by how it was previously to manually go on to change all the files ;)

meta/tasks.py
cmin764: Maybe that's why I was encountering some issues during Windows development which was accessing files on a mounted drive enabling access to files on my Mac, thing which was messing a little with my drive letters. Although not sure if `absolute()` will still break it.

Was thinking to use `resolve()` just in case we end up with `..` (parent) or `~` (home) in the path somehow.

devutils/src/devutils/invoke_utils.py
cmin764: I removed long-time ago the `## Unreleased` from every CHANGELOG because of inconsistencies and I believed that we don't want to use that.

But if it makes more sense to keep that sentinel header, then also fine by me, thus we can re-introduce it everywhere and adapt the code to work on it.

devutils/src/devutils/invoke_utils.py
OvidiuCode: projects is actually the list we pass on which is `workitems, tasks` and so on. I think it would be annoying to write the `robocorp-` suffix for all of our packages we want to install in devmode.

devutils/src/devutils/invoke_utils.py
OvidiuCode: nice catch!

devutils/src/devutils/invoke_utils.py
OvidiuCode: I don't have a strong opinion here either. We can add back the `Unreleased` header or just add the comments above the latest version. But we need to decide and be consistent.

devutils/src/devutils/invoke_utils.py
OvidiuCode: I don't think that is very important either because you can check it with a git diff and if we decide to have the `## Unrealeased` header then we don't need that piece of logic anymore.

meta/tasks.py
fabioz: > Maybe that's why I was encountering some issues during Windows development which was accessing files on a mounted drive enabling access to files on my Mac, thing which was messing a little with my drive letters. Although not sure if absolute() will still break it.

Definitely (I do have many scars on that -- `resolve()` is very seldomly needed and is usually the source of bugs). `absolute()` will do the `..` -- for `~` you need `expanduser()` (but in this case we can leave `~` if it ever gets there as there's no issue in that, the `absolute()` is only needed because if `.` was there then we could have a problem).

devutils/src/devutils/invoke_utils.py
fabioz: The good thing about `## Unreleased` -- maybe accepting any variation with any casing -- is that it gives a clear point where that'd be needed and it'd be a good point of failure.

With the current code it'd need to do more analysis to decide if the format is valid for applying the header (i.e.: it'll happily apply the header with no changes or when done a second time for the same version, having the `unreleased` would make the implementation a bit easier -- although the implementation can also be improved to handle those).

meta/tasks.py
cmin764: Looks like a strong opinion based on data, so +1 on this!

devutils/src/devutils/invoke_utils.py
cmin764: Indeed, for some reason, during testing, I thought it doesn't do the job if you pass `-l workitems` for e.g., so I fooled myself into thinking that's a bug when I tried with `-l robocorp-workitems`.

To be on the safe side, you can apply a `projects = [package.replace("robocorp-", "").strip() for package in (local or "").split(",")]` so no matter if somebody passes it with `robocorp-` prefix or not, it would still work in the normalized way.

devutils/src/devutils/invoke_utils.py
```diff
@@ -390,30 +445,63 @@ def make_release(ctx):
             echo=True,
         )
 
-        print(f"Trigger the release with: git push origin {current_tag}")
+        print(f"Pushing tag: {current_tag}")
+        run(ctx, f"git push origin {current_tag}")
```
cmin764, line 465: Post-merge, you can test this out by making a release that we really need for `excel` (which I missed to do), which will produce this tag: `robocorp-excel-0.4.3`

As currently `inv check-tag-version` in `excel`:
```
Version does not match (robocorp-excel: 0.4.3 != repo tag: 0.4.2).
Tags:['robocorp-excel-0.3.0', 'robocorp-excel-0.4.0', 'robocorp-excel-0.4.1', 'robocorp-excel-0.4.2']
(exit(1))
```

So this is a good opportunity to test the command after the Merge.

fabioz:
Hey @OvidiuCode, not sure if you missed the:

> Note: auto-generated by `_gen_workflows.py`

(in the `**tests.yml`) or not (hopefully you didn't have to manually update all the test workflows)

If you did use it, the `_gen_workflows.py` is not committed... otherwise, please do the changes you need on `_gen_workflows.py` and then make sure you run it to update things in the test workflow files.

OvidiuCode:
Thanks @fabioz. I missed that, that is why some were failing.

.github/workflows/_gen_workflows.py
```diff
@@ -160,6 +161,12 @@ def build_steps(self):
             "name": "Install poetry",
             "run": "pipx install poetry",
         }
+
+        install_devutils = {
+            "name": "Install devutils requirements",
+            "run": f"python -m pip install -r ../devutils/requirements.txt",
```
cmin764: I saw an `f""` string here but not sure what you wanted to format (as there's no variable inside). Maybe you wanted to pass in something like:
```python
"run": f"python -m pip install -r {self.devutils_requirements}",
```

.github/workflows/_gen_workflows.py
```diff
@@ -37,21 +37,22 @@ class BaseTests:
...
-        "run": "python -m invoke test",
+        "run": "inv check-all",
     }
 
     def __init__(self):
         project_dir = CURDIR.parent.parent / self.project_name
         assert project_dir.exists(), f"{project_dir} does not exist"
 
         self.target_file = CURDIR / f"{self.target}"
+        self.devutils_requirements = CURDIR / "devutils" / "requirements.txt"
```
cmin764: Won't this point out to: _.github/workflows/devutils/requirements.txt_ by mistake?

Maybe we want here:
```python
self.devutils_requirements = CURDIR.parent.parent / "devutils" / "requirements.txt"
```
as seen with `project_dir`.

.github/workflows/action_server_tests.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
         cache: poetry
-    - name: Install invoke/tomlkit
-      run: pip install invoke tomlkit
+    - name: Install devutils requirements
+      run: python -m pip install -r ../devutils/requirements.txt
```
cmin764, line 55: So the default current working directory is actually _.github_ while `CURDIR` points to _.github/workflows_, right?

Maybe we can also use a `ROOT = CURDIR.parent.parent` in case we use it often. (and `CURDIR` can refactor to `WORKFLOWS_DIR` as it is a little bit misleading when using it in the code and can trick you)

But your call here, if it works fine as it is, no must for a change.

cmin764 (commented):
Left some comments around path resolution.

If you can temporarily trigger the workflow straight from the branch just to make sure the checks & tests are passing (at least running them so we know the workflow is good), then we're good to go!

fabioz (commented):
I was just seeing the `action_server/docs/api` and I was thinking that for the action server the API isn't really helpful (as it's not a library and it's not meant to be interacted through its API, unlike the other projects in this repo).

I think we should just remove those api docs instead of updating them in this PR...

What do you think @cmin764?

cmin764:
> I was just seeing the `action_server/docs/api` and I was thinking that for the action server the API isn't really helpful (as it's not a library and it's not meant to be interacted through its API, unlike the other projects in this repo).
> 
> I think we should just remove those api docs instead of updating them in this PR...
> 
> What do you think @cmin764?

Yes, no need to generate and store API docs for the Action Server (as it's not a library on its own), and we can skip it from docs generation during the `docs` command if we detect that there's no _docs/api_ folder present already, this way fixing the `check-all` command which will also run the `docs` in CI to ensure that there's nothing newly generated.

.github/workflows/_gen_workflows.py
```diff
@@ -37,14 +37,14 @@ class BaseTests:
         ],
     }
 
-    run_tests = {
-        "name": "Test",
+    run_checks = {
```
fabioz: I was thinking a bit about this and I don't like this approach much because if there's a single check-all which does things sequentially, you'll just get the first failure, which is not ideal.

i.e.: in the previous approach it ran the tests and then it ran the lint with a `if: always()` in the clause, so, if any of those would fail both would appear as separate failures in the ci run. With the `check-all` approach one may need one ci run to see the test failures and then another ci run to see the lint failures.

My suggestion is revert to the previous approach (ideally each separate thing that could fail which is "independent" on the others should have a separate step, so, we'd have separate steps for `tests`, `lint`, `code-format`, `typecheck` and potentially `docs`).

You could also improve the check-all to do all and report separate failures even if one fails, but I still think it's easier to see in the ci each separate failure (and one does not preclude the other -- `check-all` would be something you run locally, while the ci runs them separate).

log/docs/api/robocorp.log.md
```diff
@@ -219,7 +219,7 @@ ______________________________________________________________________
 console_message(
     message: str,
     kind: str,
-    stream: Union[IO, NoneType, _SentinelUseStdout] = <robocorp.log._SentinelUseStdout object at 0x10585afb0>,
+    stream: Union[IO, NoneType, _SentinelUseStdout] = <robocorp.log._SentinelUseStdout object at 0x104fab5e0>,
```
fabioz: Humm, that'll change on each run -- maybe you can change the `_SentinelUseStdout` in this PR so that its `__repr__` and `__str__` returns `"_SentinelUseStdout"` so that it's possible to do the docs check and get stable results here?

OvidiuCode: sure can!

.github/workflows/_gen_workflows.py
cmin764: Agree, `check-all` should be something to run locally, while having all the other sub-checks run independently in CI without a fail-fast approach, thus maximizing the CI potential of explaining to you in one go what went wrong.

action_server/docs/api/README.md
cmin764, line 1: We can remove this also now @fabioz as agreed to strip of any generated API docs on non-libs.

cmin764 (commented):
LGTM!

fabioz (approved):
LGTM!

PR #246, opened 2024-02-15
Add notifiers to 3rd party library docs

### LINKED ISSUE #245: 3rd-party: notifiers

OvidiuCode:


cmin764 (approved):
If that's complete now, then this is good to go!

If you ever need more inspiration, we use this library in our [`RPA.Notifier`](https://github.com/robocorp/rpaframework/blob/master/packages/main/src/RPA/Notifier.py).

PR #248, opened 2024-02-15
Add oauthlib to 3rd party library docs

### LINKED ISSUE #247: 3rd-party: oauthlib

OvidiuCode:


cmin764 (approved):
LGTM!

If we ever need a real-world customer-tailored example, we can inspire from the legacy RPA Framework e-mail [auth](https://github.com/robocorp/rpaframework/blob/master/packages/main/src/RPA/Email/common.py) used in the following [example](https://github.com/robocorp/example-oauth-email/blob/master/tasks.robot#L108-L110).

PR #253, opened 2024-02-19
Fix READMEs for all the packages: soft-wrap and line-break

kariharju:
### This PR
- Fixes _README.md_ of every package with soft-wraps (no line width limit) and rendered line breaks (`  `) according to Markdown's author himself.
- Adds release notes for every affected package (later with [commit](https://github.com/robocorp/robocorp/commit/2a6a3249d6782a1a5a70099005881d8708447fb7)) -> next release will make the new README visible in PyPI.
### Notes
- @kariharju: The linefeeds by writer in _README.md_ files are creating a havoc on the docs -site side. When possible, use IDE's "Word Wrap" (soft-wrap) when writing Markdown to avoid manually introducing hard-wraps.
- @cmin764: At the same time, I took the chance to fix the bugs and refactor these READMEs, so they'd follow this [example](https://www.notion.so/robocorptechnologies/Markdown-fff71182e16d441b8599a40266ec9cc8?pvs=4#59d20b51b7894b7587a615b544601222) with soft-wrap and line breaks introduced by double-spaces (`  `).

### Up Next
- Make releases of every affected package, including **robocorp** (_meta_) and Action Server:
  - _actions_
  - _browser_
  - _log_ & _log_pytest_
  - _storage_
  - _tasks_
  - _vault_
  - _windows_
  - _workitems_


browser/README.md
```diff
@@ -1,15 +1,13 @@
...
 
 ## Usage
 
 ![`robocorp-browser`](https://img.shields.io/pypi/v/robocorp-browser?label=robocorp-browser)
 
-> 👉 Check that you have added the dependency in your configuration, this library is not apart of the `robocorp` -package.
-> - _conda.yaml_ for an automation [Task Packages](https://robocorp.com/docs/robot-structure)
+> 👉 Check that you have added the dependency in your configuration; this library is not a part of the `robocop` -package.
+> - _conda.yaml_ for automation [Task Packages](https://robocorp.com/docs/robot-structure)
 > - _action-package.yaml_ for an automation Action Packages
```
cmin764:
```suggestion
> - _action-package.yaml_ for automation Action Packages
```

windows/README.md
```diff
@@ -1,26 +1,22 @@
...
 
 ## Usage
 
 ![`robocorp-windows`](https://img.shields.io/pypi/v/robocorp-windows?label=robocorp-windows)
 
-> 👉 Check that you have added the dependency in your configuration, this library is not apart of the `robocorp` -package.
-> - _conda.yaml_ for an automation [Task Packages](https://robocorp.com/docs/robot-structure)
+> 👉 Check that you have added the dependency in your configuration; this library is not a part of the `robocop` -package.
+> - _conda.yaml_ for automation [Task Packages](https://robocorp.com/docs/robot-structure)
 > - _action-package.yaml_ for an automation Action Packages
```
cmin764:
```suggestion
> - _action-package.yaml_ for automation Action Packages
```

cmin764 (changes requested):
LGTM given the content, but detected two other typos. (commented below)

Now given the line length, according to `markdownlint` [docs](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length), a default line length of 80 chars is expected, but ofc, we can increase that.

What do you think @fabioz , shall we add a linter for the MDs we write/generate? Something that has a prettier preferably?

cmin764:
This may look blocked now by this issue: https://github.com/robocorp/robocorp/issues/257

But to unblock it, maybe we should manually edit these files to match what we would achieve with the solution above, then merge it as it is without staying blocked until the solution is ready.

⚠️ What I still don't understand, is why our docs site rendering part isn't compatible with the Markdown suggested [format & rules](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length) or why this started to become a problem today when we wrote line breaks in _.md_ files the entire time? Isn't this a problem that should be treated on the [**documentation**](https://github.com/robocorp/documentation) side, as long as we enforce a commonly accepted code-driven standard here in **robocorp**?

kariharju:
The length rule is not a markdown rule, just a markdown linter rule.
The docs rendering is only breaking if there are extra spaces, and the only way to get that is to manually add linefeeds.

Markdown is not code, it is content to applying old code styling rules on it really does not make sense. Markdown will in all cases (except looking at raw data in a text editor) be shown in whatever width layout so applying a some limit in the source file really does not help. 

Need to get these fixes in and let's look at the #257 separately.

cmin764:
Will refactor this PR by applying this Notion [proposal](https://www.notion.so/robocorptechnologies/Markdown-fff71182e16d441b8599a40266ec9cc8?pvs=4#855177f5d45145adafa56c36ed8a29a0).

PR #254, opened 2024-02-20
Make get_win_version a static method

### LINKED ISSUE #191: Add function to retrieve the Windows version
Add a function like `get_windows_version` which will tell you more about the Windows version you're using.

I find myself occasionally checking upon the version number in order to make a distinction between Windows 11, 10 and Server, so I can run the matching logic when I know that the UI elements and commands differ.

Examples:
- [Notepad](https://github.com/robocorp/example-windows-notepad/blob/master/tasks.py#L55-L62)
- [Calculator & Notepad](https://github.com/robocorp/example-windows-calc-notepad/blob/master/tasks.py#L47-L50)

OvidiuCode:
Make `get_win_version` a static method so it can be used more like an utilitary where needed.
Avoid redefinitions like [windows-calc-notepad](https://github.com/robocorp/example-windows-calc-notepad/blob/master/tasks.py#L32) example.

### DIFF
--- a/windows/docs/CHANGELOG.md
+++ b/windows/docs/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Make get_win_version a static method on Desktop
+- Make `get_win_version` a static method of `windows.Desktop` for easier usage.
 
 ## 1.0.2 - 2024-02-01
 
--- a/windows/docs/CHANGELOG.md
+++ b/windows/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Make get_win_version a static method on Desktop
+
 ## 1.0.2 - 2024-02-01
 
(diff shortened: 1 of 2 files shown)


windows/docs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Make get_win_version a static method on Desktop
```
cmin764:
```suggestion
- Make `get_win_version` a static method of `windows.Desktop` for easier usage.
```
Is more valuable to explain the customer what they get not just what we fixed, so they can asses from a non-technical PoV if the upgrade matters or not for them. Sometimes it helps if we get less technical in the release notes, so even a child can understand the newly brought value.

cmin764 (approved):
This looks good!

- Added one suggestion to the release note.
- Very interesting to test and observe how this function will react in Windows Server 2019 (you can give it a try and see what version will return and see if we need to fix that to match the Server case)

PR #268, opened 2024-02-28
Trigger truststore injection from tasks

OvidiuCode:
Will refactor into **browser** only injection by reusing the utility implemented with #275.


fabioz:
Note: we actually do this when running the tasks/actions in:
https://github.com/robocorp/robocorp/blob/master/tasks/src/robocorp/tasks/cli.py#L17

Do we need this to be here too? (i.e.: is someone using robocorp.browser without tasks/actions and having a problem with that?)

cmin764:
> Note: we actually do this when running the tasks/actions in: https://github.com/robocorp/robocorp/blob/master/tasks/src/robocorp/tasks/cli.py#L17
> 
> Do we need this to be here too? (i.e.: is someone using robocorp.browser without tasks/actions and having a problem with that?)

That's a good point Fabio and I was initially perplexed as well, but I think that's actually a good idea, because there's this case with the browser rcc post-install [trigger](https://github.com/robocorp/template-python-browser/pull/7/files#diff-961f789e75be01925011b708ccc8e309321a59b8fc5fe8df7e8ec27911b0a582R19) where we'd use the library entrypoint to download the browser, thing which may need to trigger the injection as it is happening **before** any task execution. And since we're going to use the network for that job, I think the injection should be in place.

Now I'm not sure at what level the injection is handled by **rcc** already if @vjmp can chime in, since I remember that was somehow needed as well for the Conda setup and PyPI packages installation, but maybe I'm wrong, as it is odd to depend on `robocorp-truststore` to download and install `robocorp-truststore`. (but if it's needed for PyPI only, then this it makes sense, as we get our **truststore** package with _conda_, not with _pip_)

browser/src/robocorp/browser/cli.py
```diff
@@ -1,10 +1,25 @@
...
+    try:
+        import truststore  # type: ignore
+
+        truststore.inject_into_ssl()
+        log.info(
+            "Truststore injection done, using system certificate store to validate HTTPS."
+        )
+    except ModuleNotFoundError:
+        log.warning("Dependency `truststore` is not installed!")
+
```
cmin764: This looks good!

Do you mind if we put this into a function in the _cli.py_ of the **tasks** package and import it from there and just call the function here? Recommending this because of the following reasons:
1. We need this change (including logging) [here](https://github.com/robocorp/robocorp/blob/master/tasks/src/robocorp/tasks/cli.py#L16-L23) as well.
2. DRY will help you here in case later on we need to come back to this and further fix/improve it. (as it is used from one single place)
    - **browser** already depends on **tasks** if you look in the _pyproject.toml_.

browser/src/robocorp/browser/cli.py
```diff
@@ -1,10 +1,25 @@
...
+# Use certificates from native storage (if `truststore` installed)
+if sys.version_info >= (3, 10):
+    try:
+        import truststore  # type: ignore
+
+        truststore.inject_into_ssl()
+        log.info(
+            "Truststore injection done, using system certificate store to validate HTTPS."
+        )
```
cmin764: While the logging does the job, do you see as an improvement if we'd try to use our own logger instead and import `robocorp.log` for that? Then use `log.info` & `log.warn` for the same, thing which might render better in the output _log.html_ file.

Or even better, adding a `log.console_message(...)` as well so we'd see it in both the console output and the log file, although this might add some duplicate messages in the output log file. Test it out and take a decision by yourself here on what you'd think it will render best!

The **robocorp-tasks** package already depends on **robocorp-log**.

cmin764 (changes requested):
The incentive is good, although maybe not complete, as I think we want to achieve the following:
1. Logging (not just injecting) in both **tasks** and **browser**.
2. Using our own special **log** library to do the logging.

Bonus: Check if the injection was done and don't second-trigger it if happened already, but most probably this improvement should go under the legacy library codebase. (so not needed in this PR)
Use-case: We have the same behaviour embedded in **rpaframework** as well, therefore if you import some `RPA.<library>` the import itself will trigger both the injection and the logging output. Maybe it will get ugly if we see different logs "duplicating" the behaviour.

browser/src/robocorp/browser/cli.py
```diff
@@ -2,6 +2,8 @@
 import argparse
 import sys
 
+from robocorp.tasks import cli  # noqa: F401
```
fabioz, line 6: Please add a comment here on why this is done -- i.e.: Importing for side effects ... .

cmin764 (approved):
Apart from @fabioz not too much to add, LGTM!

As usual, ensure you update the deps, CHANGELOG, version etc. if you're about to release afterwards.

browser/src/robocorp/browser/cli.py
cmin764: @fabioz Is this slight behaviour change (the warning issuing) now considered a new feature or just a hotfix attracting the PATCH version change?

fabioz: I'd say this is a new feature (for me, warning that the user does not have truststore is a feature, not a bugfix -- besides, users that don't want it will need to change their code to ignore the warning which is always a bit annoying).

browser/docs/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+## 2.2.4 - 2024-03-07
```
fabioz: Given that it's an "improvement", it should be `2.3.0`.

fabioz (approved):
LGTM

browser/src/robocorp/browser/cli.py
cmin764: It is indeed, just that we had too many scars with the enterprise customers having this networking issue as a blocker, so that we want all to be aware of it in case is missing.

cmin764 (commented):
Late at the party, but LGTM!

PR #275, opened 2024-03-05
SSL injection with truststore `tasks` utility

OvidiuCode:
Blocks #268 

Create `inject_truststore()` utility so we can use it in the **browser** CLI as well.


tasks/src/robocorp/tasks/__init__.py
```diff
@@ -162,6 +165,25 @@ def get_current_task() -> Optional[ITask]:
     return _task.get_current_task()
 
 
+def inject_truststore():
+    # Use certificates from native storage (if `truststore` installed)
+    if os.environ.get("RC_CHECK_TRUSTSTORE_DEP", "True") in ["false", "False", 0]:
```
cmin764: 🐛 
```suggestion
    if os.getenv("RC_CHECK_TRUSTSTORE_DEP", "True").lower() in ["false", "0"]:
```
Note that you'd always get a string from this one, so if you pass `RC_CHECK_TRUSTSTORE_DEP=0` you'd end up with `"0"`.

Also, an `os.getenv()` might look better here + lower case normalization, but your call.

----

About on the `RC_CHECK_TRUSTSTORE_DEP`, not sure if that's the right name, as this isn't just a check, but it also affects behaviour (when doing the injection), so I would rename it into `RC_USE_TRUSTSTORE` and then I would document somewhere in the docs on how the automatic injection can be disabled.

tasks/src/robocorp/tasks/__init__.py
```diff
@@ -162,6 +165,25 @@ def get_current_task() -> Optional[ITask]:
...
+        return
+
+    if sys.version_info >= (3, 10):
+        try:
+            import truststore  # type: ignore
+
+            truststore.inject_into_ssl()
+        except ModuleNotFoundError:
+            warnings.warn(
+                "Usage of the local truststore certificates can’t be enabled, ensure you have the truststore dependency in place.",
```
cmin764:
```suggestion
            warnings.warn(
                "Usage of the native system certificate stores can’t be enabled, ensure you have the `robocorp-truststore` dependency installed in the environment.",
```

tasks/src/robocorp/tasks/__init__.py
```diff
@@ -172,4 +194,5 @@ def get_current_task() -> Optional[ITask]:
     "get_current_task",
     "ITask",
     "Status",
+    "inject_truststore",
```
cmin764: Just because we plan to re-use this utility function in **browser**'s CLI as well, I wouldn't expose it to the library's public interface.

Instead I would move the `inject_truststore()` definition right under _cli.py_ (as it was before), then in **browser** will do the `from robocorp.tasks.cli import inject_truststore` so I can trigger the injection there as well symmetrically to how it was done in the **tasks**.

My main worry is this: The **tasks** library user won't ever need to import and use such `inject_truststore` during the library usage under the automation development, so we shouldn't expose that to the public interface.

Do you see any limitation with my proposed solution or clear benefit given your current one?

tasks/src/robocorp/tasks/cli.py
```diff
@@ -30,7 +20,9 @@
     sys.coinit_flags = _COINIT_MULTITHREADED  # type:ignore
 
 # Just importing is enough to register the commands
-from . import _commands  # noqa
+from . import _commands, inject_truststore  # noqa
+
+inject_truststore()
```
cmin764: Safer to keep the utility implementation here and import straight from `robocorp.tasks.cli` in **browser**. (if there's no other limitation I'm unaware of)

tasks/tests/tasks_tests/fixtures.py
```diff
@@ -8,3 +8,8 @@ def resources_dir():
     resources = Path(__file__).parent / "resources"
     assert resources.exists()
     return resources
+
+
+@pytest.fixture(autouse=True)
+def set_check_truststore_false(monkeypatch):
+    monkeypatch.setenv("RC_CHECK_TRUSTSTORE_DEP", "False")
```
cmin764: I see the intention here but I couldn't see in the review its usage, maybe you have something not pushed or unsolved yet. In any case, a better test would simulate an import error in order to trigger and capture the warning in the test, thus proving that the warning works correctly.

Additionally, you can patch the `truststore.inject_into_ssl` itself and check if it was called by default and if it wasn't called when the injection was skipped with the env var.

cmin764 (changes requested):
The high-level idea & implementation looks good to me!

I'm just worried about the place where we defined this and because it now gets available to our users, while this is an internal runtime configuration shared with other libs.

At the end, don't forget about documenting the behaviour of the env var switch somewhere in the docs and update the CHANGELOG with the upcoming release info.

tasks/src/robocorp/tasks/__init__.py
OvidiuCode: This is a good point. I think we can go with your proposal. I don't see any limitations or extra benefits from my approach.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,18 +9,36 @@
...
+
+            truststore.inject_into_ssl()
+        except ModuleNotFoundError:
+            warnings.warn(
+                "Usage of the native system certificate stores can’t be enabled, ensure you have the"
+                " `robocorp-truststore` dependency installed in the environment.",
+                Warning,
+                stacklevel=2,
+            )
+            pass
```
cmin764: Maybe there's no need for this `pass` now.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,18 +9,36 @@
...
-        truststore.inject_into_ssl()
-    except ModuleNotFoundError:
-        pass
+def inject_truststore():
+    # Use certificates from native storage (if `truststore` installed)
+
+    # RC_USE_TRUSTSTORE env var is only used to avoid unwanted warnings in tests
+    # and should not be used for other purposes
+    if os.getenv("RC_USE_TRUSTSTORE", "True").lower() in ["false", "0"]:
+        return
```
cmin764:
```suggestion
    # `_RC_TEST_USE_TRUSTSTORE` env var is only used to avoid unwanted warnings in tests
    #  and should not be used for other purposes outside of its internal usage.
    if os.getenv("_RC_TEST_USE_TRUSTSTORE", "True").lower() in ["false", "0"]:
        return
```
Added a leading underscore (`_`) to mark it as "private" and placed inside a `TEST` word to somehow mark it as required during testing.

Feel free to pick-up anything from the ideas above in case you don't want external users to start fiddle with `RC_USE_TRUSTSTORE`, but if we expect to control the behaviour publicly through this env var, then the name is already perfect.

tasks/tests/tasks_tests/fixtures.py
```diff
@@ -8,3 +8,8 @@ def resources_dir():
     resources = Path(__file__).parent / "resources"
     assert resources.exists()
     return resources
+
+
+@pytest.fixture(autouse=True)
+def set_check_truststore_false(monkeypatch):
+    monkeypatch.setenv("RC_USE_TRUSTSTORE", "False")
```
cmin764: If by any means we'll change the env var, please update it here as well.

Make sure to put a comment above explaining what the line does, something like:
> "Disables truststore SSL injection during testing so we'd avoid unnecessary warnings given the missing dependency."

cmin764 (approved):
LGTM!

But check the comments below before proceeding with the merge, as maybe you'd like a few more changes prior to that.

tasks/src/robocorp/tasks/cli.py
OvidiuCode: I like that!

PR #279, opened 2024-03-06
Whitelist actions, get package name from package.yaml and slugify package and action names in urls.

fabioz:
- The action package name is now gotten from the `package.yaml` and not from the directory name
  (it's still gotten from the directory name when `conda.yaml` is used for backward compatibility).

- The action package name and action name are slugified to be ascii only and replace
  unwanted chars for `-` in the urls.

- A `--whitelist` argument is accepted in the command line for `start` and `import` and
  it allows whitelisting action package names as well as action names.

### DIFF
--- a/action_server/docs/CHANGELOG.md
+++ b/action_server/docs/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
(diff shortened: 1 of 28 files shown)


fabioz:
Already merged.

PR #286, opened 2024-03-08
Enhance truststore injection into SSL

### LINKED ISSUE #285: Truststore SSL injection enhancement
### Target components

browser, tasks

### Why & who

Enterprise customers need to take advantage of the **robocorp-truststore** dep as early as possible (before PyPI) and they should be aware if there was any reason for the injection to not happen, while carefully replicating the behaviour we had with our battle tested `RPA.core.certificates` [logic](https://github.com/robocorp/rpaframework/blob/master/packages/core/src/RPA/core/certificates.py#L67-L68).

### Feature or enhancement

1. Currently the warning is issued when there's no `truststore` dep in place, but we should issue it when the Python/pip version is lower than the minimum accepted as well, so users will know when there's no injection happening.
2. Right now issuing Invoke commands under **tasks** will produce warnings, so we should add this dependency to the venv as well (as dev one maybe) so we won't see those warnings.

### Solution & alternative

- Check _pip_ version as well
- Issue warning when any of pip/Python versions are lower or when the dep is missing
- Ensure the dep is available in the venv (as dev dep to not interfere with the Conda one in prod)

cmin764:
Closes #285 

### ToDo

- [x] Decide if we're adding **robocorp-truststore** with minimum Py 3.10 (required) or if we drop it from here
- [x] Bump versions
- [x] Proper CHANGELOG, docs and requirements updates

### Next

- Releases: **tasks** -> **browser**


tasks/pyproject.toml
```diff
@@ -19,13 +19,14 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.9"
+python = "^3.10"
```
cmin764: This is required because of the newly added `robocorp-truststore` below.

tasks/pyproject.toml
```diff
@@ -19,13 +19,14 @@ classifiers = [
...
 robocorp-log = ">=2.4,<3"
 psutil = "^5.0"
 docstring_parser_fork = "^0.0.5"
 
 [tool.poetry.group.dev.dependencies]
 robocorp-devutils = {path = "../devutils/", develop = true}
+robocorp-truststore = "^0.8.0"
```
cmin764: Adding this here in order to avoid the warnings we'd see in the absence of such a dep during the run of the Invoke commands. (locally and maybe even in CI)

Didn't add it to the normal deps section as I don't want to have it interfering with the main **robocorp-truststore** we're bringing through Conda in the _conda.yaml_.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,35 +9,54 @@
...
 import warnings
+from importlib.metadata import PackageNotFoundError, version as get_version
 from typing import List, Optional, Protocol
 
+from packaging import version as version_parser
+
 
-def inject_truststore():
-    # Use certificates from native storage (if `truststore` installed)
+def _inject_truststore():
```
cmin764, line 28: Making it private as there's no reason/effect at all to have it publicly available.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,35 +9,54 @@
...
+    # Use certificates from the native storage. (if `truststore` installed and minimum
+    #  interpreter/pip versions are met)
 
-    # _RC_TEST_USE_TRUSTSTORE env var is only used to avoid unwanted warnings in tests
-    # and should not be used for other purposes
+    # NOTE(OvidiuCode): `_RC_TEST_USE_TRUSTSTORE` env var is only used to avoid
+    #  unwanted warnings under tests, and should not be used for other purposes.
```
cmin764, line 33: In PyCharm the contiguous comment blocks prefer some leading space. Replace `NOTE` with `TODO` / `FIXME` to observe the highlight I'm talking about, which won't happen on the following rows without the extra initial space.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,35 +9,54 @@
...
     if os.getenv("_RC_TEST_USE_TRUSTSTORE", "True").lower() in ["false", "0"]:
         return
 
-    if sys.version_info >= (3, 10):
+    try:
+        pip_version = get_version("pip")
+    except PackageNotFoundError:
+        pip_ver_ok = False
+    else:
+        pip_ver_ok = version_parser.parse(pip_version) >= version_parser.parse("23.2.1")
```
cmin764, line 42: Takes pip version into account as well as observed with the legacy battletested [behaviour](https://github.com/robocorp/rpaframework/blob/master/packages/core/src/RPA/core/certificates.py#L67-L69).

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,35 +9,54 @@
...
-                " `robocorp-truststore` dependency installed in the environment.",
-                Warning,
-                stacklevel=2,
-            )
+            pass
+        else:
+            truststore.inject_into_ssl()
+            truststore_ssl_on = True
+
+    if not truststore_ssl_on:
```
cmin764, line 54: Only when all the steps passed, we can say that we had a successful injection. Otherwise issue that warning.

tasks/src/robocorp/tasks/cli.py
```diff
@@ -9,35 +9,54 @@
...
+            " ensure you have the `robocorp-truststore` dependency installed in the"
+            " environment and that you're using Python 3.10.12 and pip 23.2.1 at"
+            " minimum",
+            Warning,
+            stacklevel=2,
+        )
 
 
-inject_truststore()
+_inject_truststore()
```
cmin764, line 65: No need to change anything in **browser**, as just importing the `robocorp.tasks.cli` will do the trigger, no matter the refactor we're having here.

cmin764 (commented):
Left self-review, please read it first before adding yours.

tasks/pyproject.toml
fabioz: I'm not sure we should still do this. See: https://sema4ai.slack.com/archives/C04SAQERGCF/p1709831458925959 for details.

tasks/pyproject.toml
fabioz: This will effectively make our minimum supported version 3.10 for our framework... so, just double-checking if this is Ok at this point.

cmin764: Yes, I think this is not ok at all, it will attract these bumps everywhere dependent, and may affect our pipelines as well in case we though to test Py3.9 (but I think we don't use a matrix of versions to test with here).

But I'm totally up to jump to Py3.10 at minimum since `truststore` doesn't work below that. We should discuss first, maybe with @mikahanninen as well in case we'll ever depend on stuff from **robocorp** as we've seen with **robocorp-storage**.

cmin764: Since `truststore` was designed to be brought by Conda, I won't interfere with, not even under our dev deps, as this will require jumping to Py3.10 which might bring more trouble than value for now.

Still, as a workaround for the missing `truststore` issuing warnings, will add it under _devutils/requirements.txt_ so the parent env out of which the Poetry ones will derive will make it available.

tasks/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp-tasks"
-version = "3.0.0"
+version = "3.0.1"
```
cmin764, line 3: Even if this looks like a breaking change now (removing the function from the public API), we agreed to mark it as a patch fix since it's affecting a file that was never meant for importing outside of our internal use and the function call won't have any effect anyway since it's already executed by the time of the import.

PR #291, opened 2024-03-12
Get into development with a single script execution

kariharju:
Scripts build the environment stated in the contribution guide and provide a _venv_, then starts VS Code inside that.

Makes sure that you have the correct Python, Poetry, Invoke etc. setups without the hassle of a global or manual setup, just get into development ASAP.

### ToDo
- [x] The bash one needs testing on Mac and appropriate refactor
- [x] Fix the Windows one to match the flow from the *nix
- [x] Update the CONTRIBUTING.md to highlight the usage of this helper script

### Caveats
- Installing with Poetry (`inv install`) afterwards won't create a new package-adjacent _.venv_ as expected, but will install dependencies in the already active _venv_ previously created.
   - Will work into this if proved to be a problem/limitation in case this script-based flow gets adopted. (not required atm)


fabioz (approved):
LGTM

devutils/develop-libraries.bat
```diff
@@ -0,0 +1,49 @@
...
+    choice /C YN /N /M "Select Y for Yes (clean environment) or N for No (use existing):"
+    IF ERRORLEVEL 2 GOTO USE_EXISTING
+)
+
+rcc venv dev-env-libraries.yaml -s dev-library-env --force
+
+::INSTALL_DEPENDENCIES
+call .\venv\Scripts\activate.bat
+
+pip install poetry==1.7 invoke==2.2
```
cmin764: I would recommend installing these from our own [requirements](https://github.com/robocorp/robocorp/blob/master/devutils/requirements.txt) list (single source of truth), which is what we also use in CI.

devutils/develop-libraries.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Check if activate script exists
+if [ -f "./venv" ]; then
+    echo "Detected existing development environment."
+    read -r -p "Do you want to create a clean environment? [Y/N] " response
+    if [[ "$response" =~ ^[Yy] ]]; then
+        echo "Creating a clean environment..."
+        ./devutils/rcc venv devutils/dev-env-libraries.yaml -s dev-library-env --force
+        ./venv/bin/activate
+        pip install poetry==1.7 invoke==2.2
```
cmin764:
```suggestion
        pip install -Ur devutils/requirements.txt
```

cmin764 (changes requested):
Asked for a small but consistent change, the one that uses the initial requirements from the file we defined already.

I like the fact that we use our own tools to setup the isolated parent venv from which all the other packages' ones derive.

devutils/develop-libraries.sh
```diff
@@ -0,0 +1,35 @@
...
+    else
+        echo "Using existing development environment."
+        ./venv/bin/activate
+    fi
+fi
+
+# Start VS Code
+code .
```
cmin764: On Mac I get: `./develop-libraries.sh: line 33: code: command not found`

Is it supposed to have this in PATH just with VSCode normally installed or shall it be done via _brew_ to have it available? (maybe manually adding it otherwise)

devutils/develop-libraries.sh
```diff
@@ -0,0 +1,35 @@
...
+if [ ! -f "./rcc" ]; then
+    curl -o rcc https://downloads.robocorp.com/rcc/releases/latest/linux64/rcc
+    chmod +x rcc
+fi
+
+# Go to repo root
+cd ..
+
+# Check if activate script exists
+if [ -f "./venv" ]; then
```
cmin764: What should happen if there's no _venv_ dir at all in place? Shouldn't we recommend to get a fresh one? (as in the existing replacing logic branch)

kariharju: The windows script builds the venv if there is none, without questions. The mac/linux script is optional, if no one needs it we do not need to add it right now.

devutils/develop-libraries.bat
kariharju: `pip>=23.2.1` in the requirements.txt is really bad, there have been multiple breaking changes to pip since 23.2.1.
23.3 was breaking and now 24 is also out. Locking these down is really crucial, especially if this affects the released libraries.

cmin764: Changed that now to match the right version we'd want to use. Thx!

devutils/develop-libraries.sh
cmin764: Fixed it meanwhile with a message in case the command isn't found or can't be executed successfully.

devutils/develop-libraries.sh
cmin764: Fixed the script to build a _venv_ in case there's no existing one at all.

PR #295, opened 2024-03-13
Show exposed URL in Action Server UI

### LINKED ISSUE #280: Show exposed URL in Action Server UI
### Target components

action_server

### Why & who

To make DX better.

### Feature or enhancement

Action Server UI has a link to OpenAPI spec, in the left side bar. When Action Server is started as exposed, it would be great to have the URL of the exposed spec available, instead of the local one. This way it would be an additional place from where to get the URL when needed.

Screenshot 2024-03-06 at 13 55 33

### Solution & alternative

_No response_

rihardsgravis:
If available, use exposed URL as the link for OpenAPI schema in Action Server UI sidebar

Closes #280


action_server/frontend/src/views/Root.tsx
```diff
@@ -185,7 +185,11 @@ const Root = () => {
               >
                 Runs
               </SideNavigation.Link>
-              <SideNavigation.Link href="/openapi.json" target="_blank" icon={<IconShare />}>
+              <SideNavigation.Link
+                href={serverConfig?.expose_url || '/openapi.json'}
+                target="_blank"
+                icon={<IconShare />}
+              >
```
fabioz: The `href` is updated, but shouldn't the `OpenAPI spec` text be changed too here?

rihardsgravis: Added the Public URL to be a separate link, so there would be less confusion:
 
image

fabioz (approved):
LGTM

PR #296, opened 2024-03-13
Display Action Run logs directly in Run View

### LINKED ISSUE #252: Provide a way to see the logs directly in the run view when it fails.
### Target components

action_server

### Why & who

@tonnitommi 
> it would be really nice to be able to access the log directly from this view where you started the run, especially when it fails

image

### Feature or enhancement

Note; the run id should've been available, so, showing a link there should be reasonably straightforward.

### Solution & alternative

_No response_

rihardsgravis:
Display Action Run log output directly in Run View:

image

Closes #252 

Note: This PR introduces also the use of `react-query` library to perform API calls. Reasoning for using this library is that it handles all the required async data loading and is battle tested as it is used in all other Robocorp UI related projects. It should also be easier to maintain for developers switching between projects.


fabioz (approved):
LGTM

PR #297, opened 2024-03-14
Improve docstrings and consistency

### LINKED ISSUE #262: Improve documentation format, layout and linting
We'll need to bring in a couple of improvements to our current docs:
1. Docstrings consistency and standard: summary, example/usage, details, arguments, errors. (and take advantage of what's already being inferred from the typing) -- expected [layout](https://linear.app/robocorp/issue/PRD-12/the-python-api-docs-layoutcontent-is-still-quite-rough)
   - Remove the redundancy of "function" / "class" and transform these into headers in the rendered docs.
   - Remove/move the code/source links at the start. (that's the least interest for the average user)
   - Every docstring should have a code example without the requirement to bring in the entire context.
3. Research and experiment with the support limitations of HTML tags. (e.g. `<br/>` in Markdowns)
4. Identify places where we use `*args` / `**kwargs` and be more explicit about what values can be passed by the customer. (at least 20% of the most frequent ones solving 80% of the problems) -- relates to #260 
5. Add docstring linters and generators (through extensions/tools if possible) that will help with crafting adequate docstrings.

### Observations

- We can do the initial experiment on a small (in size of written lines of code by ourselves) wrapper library: **browser**.
- Will affect the docstrings and [**lazydocs**](https://github.com/robocorp/lazydocs) dev.
   - Understand what it takes to continue on improving the generic **lazydocs** or building our own parser. (if we reach any challenge due to the generic nature of the parser)
- The `inv doctest` & `inv check-all` fail occasionally in many packages. (unreliable if we don't fix them soon)
- Additionally, we can add a pre-commit/push tool at the end to enforce some of the checks before pushing remotely. Might be also a good idea to include the `doctest` in our `check-all` once we have it passing in all the relevant packages.

### Related work

Other Issues/PRs which triggered the holistic work above:
- https://github.com/robocorp/robocorp/issues/257
- https://github.com/robocorp/robocorp/pull/253

OvidiuCode:
These changes are generated with the proposed changes in lazydocs draft PR: https://github.com/robocorp/lazydocs/pull/3

image


browser/src/robocorp/browser/__init__.py
```diff
@@ -25,49 +25,39 @@
...
+    Note that calling this method after the browser is initialized has no effect.
+    
+    Example:
+        ```python
+        browser.configure(browser_engine="firefox", slowmo=100)
+        ```
+    
     Args:
-        browser_engine:
-            Browser engine which should be used
```
OvidiuCode, line 36: Right now lazydocs doesn't turn this into bullet point if the description of the argument begins with new line. If we don't like it with line continuation as a standard we could look into lazydocs and change the behaviour there.

fabioz: @OvidiuCode what do you think would provide the best output format here?

OvidiuCode: I think this is very readable. I like it a lot with newline so I adjusted to code to work with newline.

fabioz: Great.

browser/src/robocorp/browser/__init__.py
```diff
@@ -280,18 +329,169 @@ def install(browser_engine: BrowserEngine, force: bool = False):
...
     from . import _engines
 
     _engines.install_browser(browser_engine, force=force)
 
 
+class BrowserConfig:
+    """
+        Browser config description here.
+
+        Put what you want in it. Testing.
```
fabioz: Don't forget to remove this class before merging ;)

browser/docs/api/README.md
cmin764, line 1: Is the removal of these permanent or will they be generated somehow before merging the PR?

browser/src/robocorp/browser/__init__.py
```diff
@@ -1,5 +1,4 @@
-"""
-Main module for doing browser automation with Playwright.
+"""Main module for doing browser automation with Playwright.
 
 This library can be made available by pinning
 ![`robocorp-browser`](https://img.shields.io/pypi/v/robocorp-browser?label=robocorp-browser)
```
cmin764: While at it, we can turn this badge into a clickable URL that will direct you to the PyPI package, I think that wrapping it with `[<content>](<url>)` might do the job, but you should try first.

browser/src/robocorp/browser/__init__.py
```diff
@@ -24,54 +23,50 @@
...
+        browser.configure(browser_engine="firefox", slowmo=100)
+        ```
+    
     Args:
-        browser_engine:
-            Browser engine which should be used
+        browser_engine: Browser engine which should be used
             default="chromium"
             choices=["chromium", "chrome", "chrome-beta", "msedge",
                      "msedge-beta", "msedge-dev", "firefox", "webkit"]
```
cmin764: Even if we split these on multiple lines, won't they get rendered as separated by spaces? Like:

> Browser engine ... used default="chromium" choices=["chromium", ...

If that happens, we'll need some extra characters/paranthesis to separate these better.

browser/src/robocorp/browser/__init__.py
```diff
@@ -24,54 +23,50 @@
...
-            Can be used to skip the playwright stop. Not recommended in general,
-            only meant to be used to diagnose and workaround specific issues on
-            the playwright stop coupled with an early os._exit shutdown in
-            `robocorp-tasks`. Can cause a process leak and even a shutdown
-            deadlock if used alone.
+            
+        skip_playwright_stop: Can be used to skip the playwright stop. Not recommended in
+            general, only meant to be used to diagnose and workaround
+            specific issues on the playwright stop coupled with an early
+            os._exit shutdown in `robocorp-tasks`. Can cause a process leak
```
cmin764:
```suggestion
            os._exit shutdown in **robocorp-tasks**. Can cause a process leak
```
In case we want to differentiate code snippets (`<code>`) from an actual **library-name**.

cmin764:
```suggestion
            `os._exit` shutdown in `robocorp-tasks`. Can cause a process leak
```
this can go as code if the construct exists.

browser/src/robocorp/browser/__init__.py
```diff
@@ -93,20 +88,19 @@ def configure(**kwargs) -> None:
...
+    for scenarios requiring different context configurations.
 
     Example:
         ```python
-        from robocorp import browser
         browser.configure_context(ignore_https_errors = True)
         ```
 
+    Args:
+        **kwargs: Keyword arguments supported by `playwright.Browser.new_context` method
```
cmin764:
```suggestion
        **kwargs: Keyword arguments supported by the `playwright.Browser.new_context` method.
```

browser/src/robocorp/browser/__init__.py
```diff
@@ -277,21 +302,147 @@ def screenshot(
...
+        isolated: bool = False,
+        skip_playwright_stop: bool = False,
+    ):
+        """Configure browser engire class.
+        
+        Args:
+            browser_engine: Browser engine which should be used
+                default="chromium"
+                choices=["chromium", "chrome", "chrome-beta", "msedge",
+                         "msedge-beta", "msedge-dev", "firefox", "webkit"]
```
cmin764: As stated above, what would be the rendering of this and are we fine with the final formatting?

As [this](https://9b5187a6.documentation-5av.pages.dev/python/robocorp/robocorp-browser/api#configure) confuses me a little bit:
Screenshot 2024-03-26 at 1 14 01 PM

We can go with something like (as the rendered Markdown):
> Browser engine which should be used. Choices: `"chromium"`, `"chrome"`, ... (defaults to `"chromium"`)

Remember, these are read by human beings as well.

cmin764 (changes requested):
Overall looks good to me and good job for the work!
Left a few improvements through the comments below.

Aside from this, it would be cool if in the future we'll find something to avoid the repetition of some `Args:` or almost identical docstrings, like deriving our functions/methods from some common interface, but no need to fret about this now.

In terms of linting these (if part of this PR or just food for thought), we should explore Ruff's [capability](https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings) of linting Google style docstrings. As the **pydocstyle** is an archived repo [pushing](https://github.com/PyCQA/pydocstyle?tab=readme-ov-file#ruff-a-modern-alternative) us to Ruff for a better and actively-maintained experience. What do you think?

Screenshot 2024-03-26 at 12 57 28 PM

browser/src/robocorp/browser/__init__.py
OvidiuCode: Agreed, will try to rephrase this

browser/src/robocorp/browser/_config.py
```diff
@@ -32,45 +32,31 @@ def __init__(
     ):
         """
         Args:
-            browser_engine:
-                Browser engine which should be used
+            browser_engine: Browser engine which should be used
                 default="chromium"
                 choices=["chromium", "chrome", "chrome-beta", "msedge",
                          "msedge-beta", "msedge-dev", "firefox", "webkit"]
```
cmin764: I see this didn't got updated as the other same parameter you refactored in `__init__.py`. Shouldn't these `Args:` be aligned with the content there?

devutils/src/devutils/invoke_utils.py
```diff
@@ -380,7 +380,6 @@ def docs(ctx, check: bool = False):
         poetry(
             ctx,
             "run lazydocs",
-            "--validate",
```
cmin764, line 383: So no validation by default, until we fix **pydocstyle** and our docstring linting configuration?

cmin764 (approved):
LGTM, gj!

Left two notes.

fabioz (approved):
LGTM

browser/docs/api/README.md
OvidiuCode: Was a problem because of the --validate param. Will have to revisit that part later.

devutils/src/devutils/invoke_utils.py
OvidiuCode: Exactly, this wasn't doing anything anyway

cmin764 (approved):
LGTM!

PR #298, opened 2024-03-18
Release libraries with `tasks` v3

cmin764:
Dependencies update for using **tasks** `3.0.0` and **meta** commands fix.

### DIFF
--- a/devutils/requirements.txt
+++ b/devutils/requirements.txt
@@ -1,4 +1,5 @@
 pip>=23.2.1
 invoke~=2.2.0
 poetry~=1.7.1
+semver~=3.0.0
 tomlkit~=0.11.8
--- a/devutils/src/devutils/invoke_utils.py
+++ b/devutils/src/devutils/invoke_utils.py
@@ -450,7 +450,8 @@ def build_common_tasks(
             print(f"No previous release for {package_name}")
         elif semver.compare(current_version, previous_version) <= 0:
             sys.stderr.write(
-                f"Current version older/same than previous: {current_version} <= {previous_version}\n"
+                f"Current version older/same than previous:"
+                f" {current_version} <= {previous_version}\n"
             )
             sys.exit(1)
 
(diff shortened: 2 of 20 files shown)


devutils/requirements.txt
```diff
@@ -1,4 +1,5 @@
 pip>=23.2.1
 invoke~=2.2.0
 poetry~=1.7.1
+semver~=3.0.0
```
cmin764, line 4: Helps running `inv update/outdated/make-release` without the need to go through a Poetry context or activate envs prior that. (as designed, since the Invoke command itself should take care of the Poetry context, but with **semver** there's an exception)

devutils/src/devutils/invoke_utils.py
```diff
@@ -487,21 +488,27 @@ def check_tag_version(ctx):
...
+                f" repo tag: {version}).\nTags:{get_all_tags(tag_prefix)}\n(exit(1))\n"
             )
             sys.exit(1)
 
     def update_changelog_file(file: Path, version: str):
         """Update the changelog file with the new version and changes"""
 
+        if not file.exists():  # usually happening with the meta one
+            print("No CHANGELOG file found, skipping update.")
+            return
```
cmin764, line 501: Needed this when ran the `inv update` command from **meta** since it errored.

meta/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "robocorp"
-version = "1.6.2"
+version = "2.0.0"
```
cmin764, line 3: Not sure if `2.0.0` is good here, as it isn't a breaking change, but attracts a major version like **tasks** `3.0.0`, so maybe it's a good idea to inline **robocorp**'s bumping with the kind of version atoms we increase under its `robocorp[-*]` related deps.

meta/tasks.py
```diff
@@ -84,8 +84,9 @@ def outdated(ctx):
 
         print(f"Checking package: {name}")
         dep_version = _fetch_version(name)
+        version = version.lstrip("^")
```
cmin764: Without this, this command would say that `^3.0.0` is different from the latest `3.0.0`.

workitems/pyproject.toml
```diff
@@ -23,7 +23,7 @@ mypy_path = "src:tests"
 
 [tool.poetry.dependencies]
 python = "^3.9"
-robocorp-tasks = ">=1,<3"
+robocorp-tasks = "^3.0.0"
```
cmin764: Strange older version pin, but anyway, we can't build now the **meta** without this upgrade, which requires a release first.

cmin764 (commented):
Self-review

meta/pyproject.toml
fabioz: `tasks` 3.0 was a breaking change (although I don't think it should affect clients, that's not impossible given that a public API changed), so, a `robocorp` `2.0` seems correct.

workitems/pyproject.toml
fabioz: As we discussed, we should probably keep the same approach with `">=1,<4"` -- nothing changed should've broken `workitems`, so, we don't really need to specify the latest as the minimum version.

PR #308, opened 2024-03-22
Fix: Add Array support to Action Server UI action run view

rihardsgravis:
- Add initial Array support to Action Server UI action run view
- Update the Action Run form to hold actual payload value types, not strings

image


action_server/frontend/src/lib/formData.ts
```diff
@@ -2,29 +2,34 @@
 import type { OpenAPIV3_1 } from 'openapi-types';
 import { InputProperty } from './types';
 
+export type PropertyFormDataType =
+  | string
+  | number
+  | boolean
+  | object
+  | Array<string | number | boolean>;
```
fabioz: Question: Does this mean that something as `Array<object>` or `Array<Array...>>` is not supported?

We can probably merge even if not (and create an issue to be handled in the future -- I guess we could even have  a "json" fallback for more complex cases).

fabioz (approved):

fabioz:
p.s.: Please add entry to changelog.md.

action_server/frontend/src/lib/formData.ts
rihardsgravis: Good catch! Fixed the type which was a bit incorrect, but `Array<object>` and `Array<Array...>>` are supported -  the only note that an `object` is not really a data type in OpenAPI – an object would always be expanded as the `properties` of that object.

For the JSON payload input I added a ticket https://github.com/robocorp/robocorp/issues/310, could add that this week at some point.

action_server/docs/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 ## 0.1.4 - 2024-03-20
 
 - Fixed issue in action-server binary build. 
+- Add support parsing Array type in Action Server UI action run view
```
fabioz: This changelog entry is actually in the wrong place (it should be in the `unreleased` section).

fabioz (approved):

PR #312, opened 2024-03-26
API Key updating issues 

### LINKED ISSUE #306: Action server react UI: API key doesn't seem to be passed to the action server when exposed.
### Target components

action_server

### What is wrong

Reported by @cmin764:

I was testing AS 0.1.4 with the [greeter](https://github.com/robocorp/actions-cookbook/blob/chore/use-robocorp-2.0.0/greeter/package.yaml) (robocorp 2.0.0) example and during the normal private/local start everything works fine, but with the public exposure I get 403 Forbidden when triggering the action through the local URL.

Although is flaky, until I change the input value, it feels like is somehow caching the previous run’s auth and no matter the changes I bring in the “API Key” field, it uses what I used last time and resets the state when I finally affect the “Name” field, see recording below:

Recording at: https://sema4ai.slack.com/files/U028FHJR5HQ/F06QMEG6NN7/screen_recording_2024-03-21_at_12.27.05___pm.mov

### System info

Windows

### Example script

_No response_

c02mo:
Fixed updating the API key & required params - running action tends to the updated values

### DIFF
--- a/action_server/frontend/.gitignore
+++ b/action_server/frontend/.gitignore
@@ -2,3 +2,5 @@ node_modules
 dist
 dist_v3
 dist-test_v3
+
+.npmrc
--- a/action_server/frontend/src/lib/requestData.ts
+++ b/action_server/frontend/src/lib/requestData.ts
@@ -2,14 +2,7 @@
 /* eslint-disable max-classes-per-file */
 
 import { Dispatch, SetStateAction } from 'react';
-import {
-  Action,
-  ActionPackage,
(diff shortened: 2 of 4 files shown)


fabioz:
LGTM.

As a note, there are some failing frontend tests which aren't really related to this PR -- it seems those are a recent addition and weren't running previously... if you could take a look at those that'd be nice (it can be in this PR or in a different one).

fabioz (approved):

PR #313, opened 2024-03-26
Format action output in local action server UI

### LINKED ISSUE #307: Format action output in local action server ui
### Target components

action_server

### Why & who

For readability of the results

### Feature or enhancement

Format the string outputs of the actions in the UI. Newlines should be newlines. This would greatly improve readability.

Screenshot 2024-03-22 at 7 21 35

### Solution & alternative

_No response_

c02mo:
Formatting result for the code mirror editor to show returns correctly. Some small stylisation to the Run drawer

- Changed 'Run' to 'Execute Action' & 'Executing' while in progress
- Added icon to button
- Formatted result string
- Added result type alongside the 'Result' header

image

image
image
image

### DIFF
--- a/action_server/frontend/src/views/actions/components/ActionRun.tsx
+++ b/action_server/frontend/src/views/actions/components/ActionRun.tsx
@@ -1,5 +1,15 @@
 import { FC, FormEvent, ReactNode, useCallback, useEffect, useState } from 'react';
-import { Box, Button, Checkbox, Form, Input, Select, Typography } from '@robocorp/components';
+import {
+  Box,
+  Button,
+  Checkbox,
+  Divider,
+  Form,
+  Input,
+  Select,
+  Typography,
+} from '@robocorp/components';
+import { IconBolt } from '@robocorp/icons/iconic';
 
 import { Action, ActionPackage } from '~/lib/types';
 import { toKebabCase } from '~/lib/helpers';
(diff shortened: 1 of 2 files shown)


action_server/frontend/src/views/actions/components/ActionRunResult.tsx
```diff
@@ -6,15 +6,24 @@ import { ActionRunConsole, Code } from '~/components';
...
+  }
+  return fmt;
+};
+
+export const ActionRunResult: FC<Props> = ({ result, runId, outputSchemaType }) => {
   return (
     <>
       <Header size="small">
-        <Header.Title title="Result" />
+        <Header.Title title={outputSchemaType ? `Result (${outputSchemaType.type})` : 'Result'} />
```
fabioz, line 24: I'm not sure about the `outputSchemaType.type` here. 
I think that on the cases the schema is simple (such as string, number, boolean) it works, but in bigger schemas it doesn't look ideal... Have you tested how it looks with a big schema (say a custom pydantic model with multiple fields)?

action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -155,12 +165,25 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
...
+          {isPending ? 'Executing...' : 'Execute Action'}
         </Button>
       </Button.Group>
-      {isSuccess && <ActionRunResult result={data.response} runId={data.runId} />}
+      <Divider />
+      {isSuccess && (
+        <ActionRunResult
+          result={data.response}
+          runId={data.runId}
+          outputSchemaType={JSON.parse(action.output_schema) as { type: string }}
```
fabioz, line 184: I think that the output schema could be an object when using a custom output model (pydantic)... I'm not sure it can be casted as string here (should be tested).

fabioz (changes requested):

c02mo:
@fabioz here is what happens when the following object is returned:
```
class Prospect(BaseModel):
    name: Annotated[str, Field(description="User's first name")]
    last_name: Annotated[str, Field(description="User's first name")]
```

image

action_server/frontend/src/views/actions/components/ActionRunResult.tsx
c02mo: Tested this. Please have a look at the top level comment.

action_server/frontend/src/views/actions/components/ActionRun.tsx
c02mo: Tested this. Please have a look at the top level comment.

fabioz (approved):

Issue #264, opened 2024-02-22
Make `robocorp-windows` develop-able on Mac/Linux as well


cmin764:
As there's no need for Windows OS to update a README, bumping the version and making a release for e.g.

Similarly to what we have in RPA Framework:
- `RPA.Windows`.[`__init__.py`](https://github.com/robocorp/rpaframework/blob/master/packages/windows/src/RPA/Windows/__init__.py#L12-L19)
- `RPA.core.windows`.[`context.py`](https://github.com/robocorp/rpaframework/blob/master/packages/core/src/RPA/core/windows/context.py#L7-L9)

PR #314, opened 2024-03-27
DEV-375 | Fixing black version in storage + improving message handling in logs

c02mo:
Fixing the raised CVE issues - updating black to patched version, added checks for properties & error handling


storage/pyproject.toml
```diff
@@ -26,6 +26,7 @@ typing-extensions = { version = "^4.6.3", python = "<3.8" }
 [tool.poetry.group.dev.dependencies]
 robocorp-devutils = { path = "../devutils/", develop = true }
 types-requests = "^2.30.0.0"
+black = { version = "^24.3.0" }
```
cmin764: I don't think we should pin **black** here, since we're not using it as a direct dependency, but as an [indirect](https://github.com/robocorp/lazydocs/blob/main/setup.py#L61) one through **lazydocs** which is coming from the _devutils/pyproject.toml_ that get's "inherited" in all our packages through this pin: `robocorp-devutils = { path = "../devutils/", develop = true }`

And I would pin it consistently like: `black = "^24.3.0"` right in _devutils_.

Then ensure I ran `inv install -u` in all our packages to make sure their _poetry.lock_ file builds and maybe even run a `inv docs && inv check-all` (after an `inv devinstall`) to ensure that these commands relying on **lazydocs** (which brings **black**) are still working (which most probably will).

----

- **lazydocs** -> dev: **black**
- **linkify-it-py** -> dev: **black**
- **typer** -> test: **black** `>=22.3.0,<23.0.0`

So hopefully that **typer** won't pose any compatibility issue during the resolution of the packages.

cmin764 (changes requested):
LGTM overall!

Requested a change on the place where this dep should be pinned and some package updates and command testing before going forward with this.

storage/pyproject.toml
fabioz: Humm, lazydocs is our own fork at this point. We can remove the 'black' dep from it (we shouldn't have a black dependency at all anywhere we control).

fabioz: One thing I don't know: it'd be nice if we wouldn't have the `dev` deps of these projects at all in our lock files. Do we know if this is possible with poetry?

i.e.: specify that we don't want these extras...

fabioz: Maybe just remove the black/lock files from this PR as that'd be a separate issue to be handled in a different PR and merge the other parts (which are unrelated to the black issue).

log/output-react/src/vscode/vscodeComm.ts
```diff
@@ -157,29 +157,57 @@ addEventListener('message', (event: any) => {
...
         // Response to something we posted.
-        const responseMsg: IResponseMessage = msg;
-        const resolvePromise = msgIdToSeq[responseMsg.request_seq];
-        if (resolvePromise) {
-          delete msgIdToSeq[responseMsg.request_seq];
-          resolvePromise(responseMsg);
+        try {
+          const responseMsg: IResponseMessage = msg;
+          if (
+            Object.hasOwn(msgIdToSeq, responseMsg.request_seq) ||
```
fabioz, line 163: Humm, looking at this now makes me want to convert the msgIdToSeq to a Map<...>

storage/pyproject.toml
c02mo: I think we should keep the scope of this PR and fix the raised CVEs.
We can do improvements afterwards.

log/output-react/src/vscode/vscodeComm.ts
c02mo: I think we should keep the scope of this PR and fix the raised CVEs.
We can do improvements afterwards.

storage/pyproject.toml
fabioz: Well, removing black would fix the CVE ;) 

But if it's trickier then we can go with the workaround until we can find out how to actually remove the dep.

tasks.py
```diff
@@ -24,12 +63,15 @@ def docs(ctx: invoke.Context) -> None:
         "meta",
     ]
 
+    if not sys.platform.lower().startswith("win"):
```
fabioz: The proper check here would be 

`if sys.platform != 'win32'`

c02mo: Sure. I can go with that.

storage/pyproject.toml
cmin764: @fabioz Yes, there's an option to Poetry install certain groups with `--only main` or `--without dev`, thus not bringing in dev deps in the committed _*.lock_ files ([documentation](https://python-poetry.org/docs/cli/#:~:text=If%20you%20want%20to%20exclude,use%20the%20%2D%2Dwithout%20option.&text=The%20%2D%2Dno%2Ddev%20option,with%20the%20%2D%2Dwith%20option.)). But we'll need to apply something like this directly into the _devutils/pyproject.toml_ source when installing the **lazydocs** from Git source:
```toml
lazydocs = { git = "https://github.com/robocorp/lazydocs.git" }
```

The odd thing is that the **black** stemming from here is an extra, meaning that this will get installed only if we install like `lazydocs[dev]`, so in the scenario of the `git` pin above, maybe there's a bug with Poetry or we have to pin by version (after a release of **lazydocs**) in order to not end up with its extra dev deps.

Even like that, **black** is brought by two other sub-deps as well:
- **linkify-it-py** -> dev: **black**
- **typer** -> test: **black** `>=22.3.0,<23.0.0`

So I'm not sure how the **black** removal from **lazydocs** will help here at all.

devutils/requirements.txt
```diff
@@ -1,6 +1,6 @@
 pip==23.2.1
 invoke~=2.2.0
-poetry~=1.7.1
+poetry~=1.8.2
```
cmin764, line 3: Why not using the latest version, so we ensure some bugs are solved already with this non-breaking bump.

devutils/src/devutils/invoke_utils.py
```diff
@@ -235,7 +235,7 @@ def devinstall(ctx, verbose: bool = False):
 
     @contextmanager
     def mark_as_develop_mode(
-        projects: Optional[list[str]] = None, all_packages: bool = False
+        projects: Optional[List[str]] = None, all_packages: bool = False
```
cmin764, line 244: I think this was to preserve compatibility with older interpreters (<3.9), thing which is still needed in **robocorp-storage** because of `python = "^3.8"` which is required by **rpaframework** since it's wrapping `robocorp.storage` while preserving compatibility with Python `>=3.8.1`.

http/tasks.py
```diff
@@ -11,4 +10,6 @@
     assert devutils_src.exists(), f"{devutils_src} does not exist!"
     sys.path.append(str(devutils_src))
 
+from devutils.invoke_utils import build_common_tasks
```
cmin764, line 13: Makes Invoke commands work again under the deprecated **http**.

devutils/.pylintrc
```diff
@@ -338,7 +338,7 @@ indent-after-paren=4
 indent-string='    '
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=88
```
cmin764, line 341: Matching what we had as default in previous black/Ruff.

tasks.py
```diff
@@ -1,36 +1,64 @@
...
 
-CURDIR: Path = Path(__file__).absolute().parent
+ROOT: Path = Path(__file__).absolute().parent
+DOCS_IGNORE = ["devutils"]
 
 
 @task
-def poetry_lock(ctx: invoke.Context) -> None:
-    """Runs 'poetry lock' on all projects with a pyproject.toml."""
+def install(
```
cmin764, line 15: Decided to combine 3 overlapping commands into one:
- `poetry-lock`, `poetry-update` -> `inv install -u`
- `poetry-install` -> `inv install`

tasks.py
```diff
@@ -1,36 +1,64 @@
...
 def docs(ctx: invoke.Context) -> None:
     """Regenerate documentation for each library."""
-    ignored = [
-        "devutils",
-        "integration_tests",
-        "meta",
-    ]
+    ignored = DOCS_IGNORE[:]
+    if sys.platform != "win32":
+        ignored.append("windows")
```
cmin764, line 53: This won't be required once we enable **windows** dev on Mac/Linux for simple tasks, like docs generation. #264

cmin764 (approved):
Added some of my changes as well, meanwhile, Dependabot just [closed](https://github.com/robocorp/robocorp/pull/305#issuecomment-2022875297) its own PR since **black** was detected to not be present anymore among its deps, so it was a temporary issue with how the outdated _*.lock_ files were generated.

Screenshot 2024-04-01 at 1 29 02 PM

Now this **black** package isn't listed as a dep anymore in any of the _poetry.lock_ files. (just referenced as a dev/test dep)

cmin764:
I'm merging this now since I'm blocked by it due to other upcoming development. And if we don't like something post-merge, we can revisit and deliver more value into this area.

PR #321, opened 2024-03-28
Update documentation api with the latest format

### LINKED ISSUE #262: Improve documentation format, layout and linting

OvidiuCode:
Use the latest lazydocs improvements to generate documentation.

### DIFF
--- a/actions/docs/api/robocorp.actions.cli.md
+++ b/actions/docs/api/robocorp.actions.cli.md
@@ -2,21 +2,11 @@
 
 # module `robocorp.actions.cli`
 
-**Source:** [`cli.py:0`](https://github.com/robocorp/robocorp/tree/master/actions/src/robocorp/actions/cli.py#L0)
+# Functions
 
 ______________________________________________________________________
 
-## function `main`
-
-**Source:** [`cli.py:12`](https://github.com/robocorp/robocorp/tree/master/actions/src/robocorp/actions/cli.py#L12)
-
-```python
-main(
-    args=None,
-    exit: bool = True,
-    plugin_manager: Optional[ForwardRef('_PluginManager')] = None
-) → int
-```
+## `main`
 
 Entry point for running actions from robocorp-actions.
 
@@ -26,7 +16,17 @@ Entry point for running actions from robocorp-actions.
 
 - <b>`exit`</b>:  Determines if the process should exit right after executing the command.
 
-plugin_manager: Provides a way to customize internal functionality (should not be used by external clients in general).
+- <b>`plugin_manager`</b>:  Provides a way to customize internal functionality (should not be used by external clients in general).
 
 **Returns:**
 The exit code for the process.
+
+[**Link to source**](https://github.com/robocorp/robocorp/tree/master/actions/src/robocorp/actions/cli.py#L12)
+
+```python
+main(
+    args=None,
+    exit: bool = True,
+    plugin_manager: Optional[ForwardRef('_PluginManager')] = None
+) → int
+```
--- a/actions/docs/api/robocorp.actions.md
+++ b/actions/docs/api/robocorp.actions.md
@@ -2,17 +2,11 @@
 
 # module `robocorp.actions`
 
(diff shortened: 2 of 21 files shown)


cmin764:
```bash
% git diff HEAD
diff --git a/windows/docs/api/robocorp.windows.md b/windows/docs/api/robocorp.windows.md
index 0529bd7..5576e2e 100755
--- a/windows/docs/api/robocorp.windows.md
+++ b/windows/docs/api/robocorp.windows.md
@@ -1272,7 +1272,7 @@ set_value(
     enter: bool = False,
     newline: bool = False,
     send_keys_fallback: bool = True,
-    validator: Optional[Callable] = <function set_value_validator at 0x000001949127DC60>,
+    validator: Optional[Callable] = <function set_value_validator at 0x00000168614A6950>,^M
     locator: Optional[str] = None,
     search_depth: int = 8,
     timeout: Optional[float] = None
```

The flaky problem with the _windows_ docs generation might be solved with such decorator if you don't find anything else simpler:

```python
class function_signature:
    def __init__(self, value):
        self.value = value
    def __call__(self, func):
        class wrapper:
            def __init__(this, decorator):
                this._decorator = decorator
            @functools.wraps(func)
            def __call__(this, *args, **kwargs):
                return func(*args, **kwargs)
            def __str__(this):
                return this._decorator.value
            def __repr__(this):
                return this._decorator.value
        return wrapper(self)
    
@function_signature("this is my function")
def my_function(a: int = 10):
    """Docstring."""
    print(a)
    
>>> my_function()
10
>>> my_function
this is my function
>>> repr(my_function)
'this is my function'
```

With the caveat that `help(my_function)` won't render so nicely:
```python
class wrapper(builtins.object)
 |  wrapper(decorator)
 |  
 |  Methods defined here:
 |  
 |  __call__ = my_function(a: int = 10)
 |      Docstring.
 |  
 |  __init__(this, decorator)
 |  
 |  __repr__(this)
 |  
 |  __str__(this)
```

cmin764 (approved):
This LGTM!

But before we go with this massive change, since we are at it, don't we want to place the "Link to source" part at the very end (right after the function signature)? It looks very odd to me to have a call for action that will send me on another page, then still have some more details about the function I was already reading:

Screenshot_2024-03-28_at_5_07_34 PM

cc @kariharju

PR #322, opened 2024-03-28
Verify if there is an action-server newer version

### LINKED ISSUE #187: Verify if the current version is the latest when starting up.
Something like:

```
A new release of action-server is available: 2.40.1 → 2.42.1
To upgrade, run: ...
```

Note: need to detect how it was installed to provide instructions on how to upgrade.

c02mo:
Record on how it looks in the CLI when an update is found:

image

if no newer version, no text is shown.

### DIFF
--- a/action_server/go-wrapper/main.go
+++ b/action_server/go-wrapper/main.go
@@ -3,15 +3,120 @@ package main
 import (
 	"embed"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 )
 
 //go:embed all:assets/*
 var content embed.FS
+
+// Constants
+const VERSION_LATEST_URL = "https://downloads.robocorp.com/action-server/releases/latest/version.txt"
+const ACTION_SERVER_LATEST_BASE_URL = "https://downloads.robocorp.com/action-server/releases/latest/"
+
+func getLatestVersion() (string, error) {
+	// Get the data from the URL
+	versionResponse, err := http.Get(VERSION_LATEST_URL)
+	if err != nil {
+		return "", err
+	}
+	defer versionResponse.Body.Close()
+
+	// Read the body content
+	versionInBytes, err := io.ReadAll(versionResponse.Body)
+	if err != nil {
+		return "", err
+	}
+	// Convert the byte slice to string
+	versionAsString := string(versionInBytes)
+	return strings.TrimSpace(versionAsString), nil
+}
+
+func parseVersion(version string) ([]int, error) {
+	versionParts := strings.Split(version, ".")
+	parsed := make([]int, len(versionParts))
+	for i, part := range versionParts {
+		num, err := strconv.Atoi(part)
+		if err != nil {
+			return make([]int, 0), fmt.Errorf("invalid semantic version format")
+		}
+		parsed[i] = num
+	}
+	return parsed, nil
+}
+
+func compareSemanticVersions(v1 string, v2 string) (int, error) {
+	// Parse given versions
+	v1Parsed, err1 := parseVersion(v1)
+	if err1 != nil {
+		return -2, err1
+	}
+	v2Parsed, err2 := parseVersion(v2)
+	if err2 != nil {
+		return -2, err2
+	}
+
+	// Compare versions
+	if len(v1Parsed) != len(v2Parsed) {
+		return -2, fmt.Errorf("incompatible semantic versions found: '%s' or '%s'", v1, v2)
+	}
+
+	for i, v := range v1Parsed {
+		if v != v2Parsed[i] {
+			if v < v2Parsed[i] {
+				return -1, nil
+			} else {
+				return 1, nil
+			}
+		}
+	}
+	return 0, nil
+}
+
+func checkAvailableUpdate(version string) {
+	// Get the latest version
+	latestVersion, err := getLatestVersion()
+	if err != nil {
+		fmt.Println("Verifying latest version failed:", err)
+		return
+	}
+	// Compare the given version with the latest one
+	compareResult, err := compareSemanticVersions(version, latestVersion)
+	if err != nil {
+		fmt.Println("An error occurred while comparing versions:", err)
+		return
+	}
+
+	switch compareResult {
+	case -1:
+		// Construct the needed URL path to get to the downloadable object
+		var actionOS, actionExe string
+		switch runtime.GOOS {
+		case "windows":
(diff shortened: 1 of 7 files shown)


action_server/src/robocorp/action_server/cli.py
```diff
@@ -333,7 +333,7 @@ def _create_parser():
     return base_parser
 
 
-def main(args: Optional[list[str]] = None, *, exit=True) -> int:
+def main(args: Optional[List[str]] = None, *, exit=True) -> int:
```
fabioz: The action server itself is only Python 3.10 onwards (as it's not a library used by clients and is meant to be used as an executable we don't need to use older patterns -- meaning `list[str]` is preferred over `List[str]`).

action_server/go-wrapper/main.go
```diff
@@ -3,16 +3,121 @@ package main
...
+		case "darwin":
+			actionOS = "macos64"
+			actionExe = "action-server"
+		default:
+			fmt.Println("Unsupported operating system")
+			os.Exit(1)
+		}
+		urlPath, _ := url.JoinPath(ACTION_SERVER_LATEST_BASE_URL, actionOS, actionExe)
+		fmt.Printf("\n ⏫ A new version of action-server is now available: %s → %s \n", version, latestVersion)
+		fmt.Printf("    To update, download from: %s \n\n", urlPath)
```
fabioz: Looks good to me, but I wonder if users are getting it from some other source (for instance I think we guide users to use `brew`), so, just double check with @kariharju if printing as is is a good option in Mac OS.

action_server/go-wrapper/main.go
```diff
@@ -3,16 +3,121 @@ package main
...
+	"strconv"
 	"strings"
 )
 
 //go:embed all:assets/*
 var content embed.FS
 
+// Constants
+const VERSION_LATEST_URL = "https://downloads.robocorp.com/action-server/releases/latest/version.txt"
+const ACTION_SERVER_LATEST_BASE_URL = "https://downloads.robocorp.com/action-server/releases/latest/"
```
bogdancondurache: I would construct the VERSION_LATEST_URL variable from ACTION_SERVER_LATEST_BASE_URL

action_server/go-wrapper/main.go
```diff
@@ -3,16 +3,121 @@ package main
...
+	versionInBytes, err := io.ReadAll(versionResponse.Body)
+	if err != nil {
+		return "", err
+	}
+	// Convert the byte slice to string
+	versionAsString := string(versionInBytes)
+	return strings.TrimSpace(versionAsString), nil
+}
+
+func parseVersion(version string) ([]int, error) {
```
bogdancondurache: I would stay with strings vs strings because our versions can contain `-beta` and this will break that functionality

action_server/src/robocorp/action_server/cli.py
c02mo: OK. Will revert.

action_server/go-wrapper/main.go
c02mo: I can add that suggestion as well only for darwin systems.

bogdancondurache (changes requested):

action_server/go-wrapper/main.go
```diff
@@ -3,16 +3,126 @@ package main
...
+	if err1 != nil {
+		return -2, err1
+	}
+	v2Parsed, err2 := parseVersion(v2)
+	if err2 != nil {
+		return -2, err2
+	}
+
+	// Compare versions
+	if len(v1Parsed) != len(v2Parsed) {
```
bogdancondurache: this seems a bit tricky if not documented, specially if using str compare, maybe it's easier to just return bool and use the error?

action_server/tasks.py
```diff
@@ -101,6 +117,18 @@ def build_frontend(ctx: Context, debug: bool = False, install: bool = True):
...
+@task
+def build_binary(ctx: Context) -> None:
+    with _change_to_build_binary():
+        run(ctx, "pyoxidizer", "run", "--release")
+
+
+@task
+def build_go_wrapper(ctx: Context) -> None:
+    with _change_to_build_go_wrapper():
+        run(ctx, "go", "build", "-o", "action-server-unsigned")
```
bogdancondurache, line 129: just out of curiosity, was this tested, if it works? I'm asking because it's quite tricky to get this right as files have to be in the right place etc. so probably the wrapper won't actually contain the AS build. Or if it works... I would be really curious how 😄

action_server/go-wrapper/main.go
c02mo: Sure

action_server/go-wrapper/main.go
c02mo: Hmmm... That's not indicated anywhere, first of all.
Secondly, comparing strings in this situation would give you only if there is a difference between them and not if there is an upgrade or not. 
If we do use the additional forms of the semver than it's probably better to use an already built package for it. I'll do that.

action_server/go-wrapper/main.go
c02mo: Probably will be resolved by the other comment. 
Switching to a package.

action_server/tasks.py
c02mo: Yes. I've tested this. The `assets` folder seems to be handled correctly by the binary.
Give me an example of what I can test locally to make sure this works.

action_server/go-wrapper/main.go
```diff
@@ -3,16 +3,82 @@ package main
...
+	latestVersion, err := getLatestVersion()
+	if err != nil {
+		fmt.Println("Verifying latest version failed:", err)
+		return
+	}
+	// Compare the given version with the latest one
+	compareResult := semver.Compare(strings.TrimSpace("v"+version), strings.TrimSpace("v"+latestVersion))
+
+	switch compareResult {
+	case -1:
```
bogdancondurache: is this needed to be a switch if the only checked case is -1? Also this seems to be a case where early returns would look better as in:

```go
if compareResult != -1 {
    return
}
var actionOS, actionExe string
.
.
.
```

action_server/tasks.py
bogdancondurache: oh, did not exact that, awesome then!
And not needed, if building and running works, that's perfect

action_server/go-wrapper/main.go
c02mo: Good point. Something that was kept for the other compare results.
Changed it.

bogdancondurache (approved):
looks good, ship it 👍

PR #323, opened 2024-03-28
Add support for direct JSON payload input in Action Server UI run view

### LINKED ISSUE #310: Add support for direct JSON payload input in Action Server UI run view
### Target components

action_server

### Why & who

For ease of use 

### Feature or enhancement

Direct raw JSON input should be supported in the Action Server run UI as an alternative to the form input. Would help a lot in ease of use when working with complex payloads

### Solution & alternative

_No response_

c02mo:
image

image


action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -64,6 +77,64 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
...
+          }
+        });
+
+        debounce(() => setFormData([...tempForm]), 750)();
+      } catch (e) {
+        setErrorJSON('Validating JSON syntax failed. Please verify the input and try again.');
+      }
+    },
+    [formData],
+  );
```
rihardsgravis: This approach, would not work with nested items (which are flattened in the `formData`), which are supported by the spec.

rihardsgravis (commented):
I would not go into the lengths of updating `formData` with the JSON payload, since that approach is also broken in the current implementation, but just:
- Send the JSON "as is" if the json mode is active (the user is responsible for writting it valid)
- And if switched back to form mode – the form would be just empty

action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -64,6 +77,64 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
...
+          default:
+            output[prop.name.toLowerCase()] = prop.value;
+            break;
+        }
+      });
+      return JSON.stringify(output, null, 4);
+    } catch (e) {
+      return 'There was an error while parsing form data...';
+    }
+  }, [formData]);
```
rihardsgravis: The JSON should be created in a nested manner – `FormData` is in a flattened form.
There is already a helper function that can be used for this at https://github.com/robocorp/robocorp/blob/f98345ef852a863a2fbb36b279086c2138700d41/action_server/frontend/src/lib/formData.ts#L133

rihardsgravis (changes requested):

action_server/frontend/src/views/actions/components/ActionRun.tsx
c02mo: Seen that. Fixed it.

action_server/frontend/src/views/actions/components/ActionRun.tsx
c02mo: Done.

c02mo:
@rihardsgravis @fabioz Can you take another look over this so we can push it?

action_server/frontend/src/lib/formData.ts
```diff
@@ -174,3 +174,42 @@ export const formDatatoPayload = (data: PropertyFormData[]): Payload => {
 
   return result;
 };
+
+export const payloadToFormData = (
```
rihardsgravis, line 178: This would be a great candidate for a unit test. For an example, see [formDataToPayload](https://github.com/robocorp/robocorp/blob/master/action_server/frontend/src/__tests/formData.test.ts) test file.

action_server/frontend/src/lib/helpers.ts
```diff
@@ -55,3 +55,13 @@ export const prettyPrint = (input: string) => {
...
+
+let debounceTimer: NodeJS.Timeout;
+export const debounce = (func: (...data: unknown[]) => void, timeout = 300) => {
+  return (...args: unknown[]) => {
+    clearTimeout(debounceTimer);
+    debounceTimer = setTimeout(() => {
+      func.apply(this, args);
+    }, timeout);
+  };
+};
```
rihardsgravis, line 67: If not used, let's maybe remove it?
```suggestion
```

action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -128,13 +132,45 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
...
+    }
+    const payload = formDataToPayload(formData);
+    setFormRawJSON(JSON.stringify(payload, null, 4));
+  }, [formData]);
+
+  useEffect(() => {
+    if (!useRawJSON && formRawJSON && formData.length >= 0) {
+      setFormData(payloadToFormData(JSON.parse(formRawJSON), formData));
+    }
+  }, [useRawJSON]);
```
rihardsgravis, line 172: I think `setFormRawJSON` can be affected by `useRawJSON` not on every formData change,  so both of these effects can be merged:

```suggestion
  useEffect(() => {
    if (!useRawJSON){
      setFormData(payloadToFormData(JSON.parse(formRawJSON), formData));
    } else {
      const payload = formDataToPayload(formData);
      setFormRawJSON(JSON.stringify(payload, null, 4));
    }
  }, [useRawJSON]);
```

action_server/frontend/src/components/Code.tsx
```diff
@@ -60,15 +62,16 @@ export const Code: FC<CodeProps> = ({
   }, [autoFocus]);
 
   return (
-    <Box height={height} mb="$24">
+    <Box key="base-code-input-wrapper" height={height} mb="$24">
```
rihardsgravis, line 65: No need/effect for a key in a single item component
```suggestion
    <Box height={height} mb="$24">
```

rihardsgravis (approved):
LGTM 👏  Some suggestions only.

fabioz:
This is now merged in https://github.com/robocorp/robocorp/commit/a9c8370ac95c6f21eaac3fd3b9878e132001702d

PR #324, opened 2024-03-29
Add sentinel class to generate consistent docs

OvidiuCode:


windows/docs/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Stable documentation generation
```
cmin764:
```suggestion
- Stable API documentation generation impacting parameters taking functions as defaults.
```

windows/docs/CHANGELOG.md
cmin764, line 1: There's a merge conflict on this now. (check latest `master`)

windows/src/robocorp/windows/_control_element.py
```diff
@@ -30,6 +30,17 @@
 DEFAULT_SEND_KEYS_INTERVAL = 0.01
 
 
+class _SentinelValidator:
```
cmin764, line 33: Doesn't hurt to have a comment explaining that this is used to be replaced with a default function when `validator` param's value stays unchanged.

cmin764 (approved):
LGTM! (couple of easy-fix light comments)

windows/src/robocorp/windows/_control_element.py
```diff
@@ -30,6 +30,22 @@
...
+    """
+    Used to determine the validator function for ControlElement.set_value method.
+    If a validator is not passed then set_value_validator will be used as default.
+    """
```
cmin764:
```suggestion
    """Used to determine the validator function for `ControlElement.set_value` method.
    
    If a validator is not passed then the `set_value_validator` function will be used as default.
    """
```
This is how it should be to match our latest docstring rules.

But why having a docstring here when it should have been a comment? (since `_SentinelValidator` belongs to the private API, which can be documented through comments, not docstrings)

cmin764 (approved):
LGTM! (just that maybe we want a comment instead of a docstring according to [this](https://www.notion.so/robocorptechnologies/Docstrings-c41098ba80cf4ba79036562aba6772b7?pvs=4#8cfd56c066ca4a31af928d50c290f87d))

cmin764:
~But for some reason, I get this now out of the blue, no matter the venv approach I'm using (Pyenv with Poetry's _.venv_ or our RCC _venv_):~
```bat
> inv docs
poetry run lazydocs --no-watermark --remove-package-prefix --src-base-url https://github.com/robocorp/robocorp/tree/master/ --overview-file README.md --output-path Z:\Repos\robocorp\windows\docs\api robocorp.windows
Failed to generate markdown for robocorp.windows.
```

Solved now! (`inv install -u` also installs)

PR #329, opened 2024-04-03
Enable Markdown linter support and EOL normalization

### LINKED ISSUE #257: Markdown linting and EOLs consistency
### Target components

devutils, docs

### Why & who

Internal/external devs, ReMark and the documentation website which has some trouble rendering different ways of splitting the lines when parsing our _.md_ sources.

### Feature or enhancement

- There's no New Line consistency enforced. (`\n` mixed with `\r\n`)
- Spacing is inconsistent. (some files ending with 0, 1 or 2 new lines; trailing spaces on non-empty lines; useless empty lines)
- Everything relies on human opinions, place where we need a tool to automate the prettying and enforce a standard through linting.

### Solution & alternative

1. All IDEs we use should strip trailing spaces and end the file with only one mandatory NL. (same standard for Python code) — easily configurable in IDE’s settings
2. A Markdown [prettier](https://github.com/prettier/prettier) and [linter](https://github.com/markdownlint/markdownlint) should be automated in our Invoke command stack. ([extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint))
3. Git [attributes](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings) configuration to [normalize](https://stackoverflow.com/a/50645024) the line endings, so we also stay consistent with what we end-up in the repo vs. what everyone has locally.

Slack public [thread](https://github.com/robocorp/robocorp/issues/262) -- [Linear](https://linear.app/robocorp/issue/DAP-79/the-python-api-docs-layoutcontent-is-still-quite-rough)
Slack private [thread](https://robocorptechnologies.slack.com/archives/C063TMX3R7Y/p1708446936457469)

cmin764:
Closes #257 

This PR adds support for Markdown lint tool during the `inv lint` command, only if the _mdl_ executable is available in PATH.

Beside, this takes advantage of some other important opportunistic refactors:
- Introduction of _.gitattributes_ so we'd stay consistent with EOLs on all the files in the repo.
- Markdown lint tool configuration automatically used with our `inv lint` command.
- `excel`: Fixes main README and regenerated docs for PyPI clickable badge.
- `windows`:  Regenerated docs due to EOL normalization (`\n`) and PyPI clickable badge.

### Up next (optional)

- Run the linter in every package and ensure necessary _*.md_ refactors. (or enable a [prettier](https://github.com/prettier/prettier) for it if the installation can be automated with ease)
- Make our own Markdown lint [Rules](https://github.com/markdownlint/markdownlint/blob/main/docs/creating_rules.md) which will check if the lines in a paragraph are correctly separated by a [line break](https://www.notion.so/robocorptechnologies/Markdown-fff71182e16d441b8599a40266ec9cc8?pvs=4#2c4253fd4a274aac961179977ca16051).


.gitattributes
cmin764, line 1: For Markdown files, enforce _LF_ (Mac/Linux), for everything else which is **text**, use Git's best knowledge (which defaults to the majority, being _LF_ as well).

browser/README.md
cmin764, line 1: With these changes, the Markdown linter doesn't scream anymore. (but we still need to apply it for the rest of the _*.md_ files in the package and in other packages as well in a different PR)

devutils/src/devutils/invoke_utils.py
```diff
@@ -277,15 +278,55 @@ def mark_as_develop_mode(
...
+        if resolved_path:
+            print("Running Markdown lint over the hand-written docs...")
+            docs_paths = " ".join(map(str, _collect_docs_to_lint()))
+            cmd = f"{resolved_path} --config {ROOT / '.mdlrc'} {docs_paths}"
+            subprocess.check_call(shlex.split(cmd))
+        else:
+            print(
+                "Markdown lint wasn't found in PATH, consider installing and adding it"
+                " in order to be able to lint package's *.md documents."
+            )
```
cmin764, line 328: Couldn't install it by _brew_ nor _pip_. Can't even imagine what it takes on Windows (chocolatey + Ruby + gem).

But, even if this would be dead unused code for you, it's the first stepping stone into enabling Markdown linting that really works without producing a dozen of errors.

Next (different PR) would be to find a way of automatically installing the linter executable, in CI as well.

excel/src/robocorp/excel/__init__.py
```diff
@@ -2,7 +2,7 @@
 Main module for doing Excel automation.
 
 This library can be made available by pinning
-![`robocorp-excel`](https://img.shields.io/pypi/v/robocorp-excel?label=robocorp-excel)
+[![`robocorp-excel`](https://img.shields.io/pypi/v/robocorp-excel?label=robocorp-excel)](https://pypi.org/project/robocorp-excel/)
```
cmin764, line 5: We want these clickable anywhere in the handwritten or generated docs. (staying consistent with the rest of the pins)

windows/docs/api/README.md
cmin764, line 1: Very big diffs upcoming on the **windows** package due to EOLs normalization. (_CRLF_ -> _LF_)

But the code stays the same, only the line endings are changing.

cmin764 (commented):
Self-review

windows/docs/api/README.md
cmin764: @fabioz Please pull and test this branch on your Windows machine, as me & @OvidiuCode are using Macs, therefore I want to know if your IDE/workflow suffers any change/burden due to these. (in theory, these should go unnoticed to you as well)

OvidiuCode (approved):

.gitattributes
fabioz: I have a question: why use `lf` for markdown instead of using `text` as is done for `.py`/`.js`?

cmin764: You have a point, as I wasn't sure about what I am doing:
Screenshot 2024-04-04 at 8 44 46 PM

So will refactor as you suggested, to treat everything equally.

The only explicit setting of either _LF_ or _CRLF_ would be when somebody will start having trouble with the line endings in their IDE, so we switch to something that works for everybody, no matter the OS you're developing. Thx!

PR #336, opened 2024-04-05
Add docstrings linter via robocorp pydocstyle

OvidiuCode:
Depends on: https://github.com/robocorp/pydocstyle/pull/2 and https://github.com/robocorp/lazydocs/pull/4


fabioz (changes requested):
This PR should also refresh all the lock files.

You can add:

```
@task
def lock(
    ctx: invoke.Context,
    skip: Optional[list[str]] = None,
) -> None:
    """Make lock for all Poetry-based projects' environments.

    Args:
        skip: List of project directories to skip.
    """
    for project_dir in _iter_project_dirs():
        project_name = project_dir.name
        if skip and project_name in skip:
            print(f"Skipping project {project_name!r}.")
            continue

        subprocess.check_call(["poetry", "lock"], cwd=project_dir)
```

to the `tasks.py` in the root and run `inv lock` to do that.

fabioz (approved):
LGTM

Issue #344, opened 2024-04-10
Action Server UI is not handling int fields properly


fabioz:
### Target components

action_server

### What is wrong

Report by @mikahanninen:

Having something as:

```
@action
def get_top_links(url: str, count: int) -> Links:
    """Gets the top links of a website

    Args:
        url (str): URL of the website
        count (int): Count on how many links to retrieve

    Returns:
        Links: Top links of the website
    """
```
In the Action Server run view, the Count field is zero and I can't change it to anything else

### System info

Problem in the action server UI (Windows).

fabioz:
Fixed in https://github.com/robocorp/robocorp/commit/ef7e6caea80cfdd2a55563d335d38608a63542c5

PR #345, opened 2024-04-11
Fix issue with number inputs on Action Server run UI

rihardsgravis:
Fixes issue #344 where in the Action Server run view, the Count field is zero and the user can't change it.


fabioz:
LGTM

fabioz:
Fixed in https://github.com/robocorp/robocorp/commit/ef7e6caea80cfdd2a55563d335d38608a63542c5

Issue #342, opened 2024-04-09
Package metadata output "info.title" should be set to package name


ktaube:
### Target components

action_server

### What is wrong

Running `action-server package metadata` should return `openapi.json` where `info.title` field is set to the package name from `package.yml`

### System info

`action-server@0.2.1`

PR #367, opened 2024-04-15
Expose the action package description in package metadata

ktaube:
Related to #342 

Continuation of https://github.com/robocorp/robocorp/pull/366.

We currently have no reliable way to tell the action package description from the `metadata`.

- Adds `metadata["metadata"]["description"]` field

e.g. for `package.yaml`
```yaml

description: Kristaps dev actions

...
```
you'd get
```json
{
  "metadata": {
    "secrets": {},
    "description": "Kristaps dev actions"
  },
  "openapi.json": {
  // ...
  }
}
```

### DIFF
--- a/action_server/docs/CHANGELOG.md
+++ b/action_server/docs/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Return package name in package metadata
+- Return package description in package metadata
 
 ## 0.3.2 - 2024-04-12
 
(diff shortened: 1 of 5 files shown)


fabioz (changes requested):
Hey @ktaube , I think for this particular use-case you should be able to just load the `package_dir/package.yaml` and collect the needed info for it (so, the db shouldn't need to be changed for that).

ktaube:
> Hey @ktaube , I think for this particular use-case you should be able to just load the `package_dir/package.yaml` and collect the needed info for it (so, the db shouldn't need to be changed for that).

Hey, got it - PR updated!

PR #371, opened 2024-04-16
Argument names always shown, added ` (item)` to the fields related to list items. Fixes #370

### LINKED ISSUE #370: Action Server UI does not always show the argument names
### Target components

action_server

### What is wrong

The input argument names are not always visible in the action UI.

1. Take the template action with `action-server new`
2. `action-server start`
3. Open the Action Server UI
4. Go to run the action
-> The argument names `user_timezone` and `compare_to_timezones` are not visible.

There is something random at play here because this state was somehow triggered by adding default values for the input arguments. After running that once the UI stopped showing the arg names.
The other image below shows a case where only the first argument name is missing.

image
image

### System info

Happened on Windows and mac.
Template action project
Action Server v0.3.2

### Example script

_No response_

fabioz:


action_server/frontend/src/views/actions/components/ActionRun.tsx
```diff
@@ -244,82 +229,70 @@ export const ActionRun: FC<Props> = ({ action, actionPackage }) => {
...
-                    <Checkbox
-                      key={`${title}-${item.name}-${index}`}
-                      label={title}
-                      description={item.property.description}
-                      checked={typeof item.value === 'boolean' && item.value}
-                      required={item.required}
-                      onChange={(e) => handleInputChange(e.target.checked, index)}
-                    />
-                  </Item>
+                  <Checkbox
```
fabioz: Note: the issue fixed here is that `item.title` isn't really defined, but after fixing then it was shown twice (one in `<Item title` and another in `<Checkbox label`, `<Input.label`, etc).

So, I opted to just remove the `Item` part.

action_server/frontend/src/lib/formData.ts
```diff
@@ -141,6 +142,24 @@ export const propertiesToFormData = (
...
 };
 
+/**
+ * Sets the title for one item of a list (i.e.: each item of a list
+ * should not have the same title as the list itself so that it's
+ * clearer what's an item of a list vs the list itself).
+ * 
+ * @param item The item which should have the title set.
+ */
+export const setArrayItemTitle = (item: PropertyFormData) => {
```
fabioz: This is an additional fix to list items so it's clearer what's the `list` item and what's the `list (item)`.

Final UI:

image

rihardsgravis:
With this change all the items loose indentation and for nested object values it is really hard to understand what item belongs to what 🤔 Also the nested object now loose the title of the parent object.

rihardsgravis:
I think the only change required was to change `label={item.title}` to `label={title}` for the `Input` of type "string"?

fabioz:
Agreed (on both). Will update the PR.

rihardsgravis (changes requested):
I would revert the changes (or update the logic to support) as they break the formatting of nested items and nested object loose the information of the parent, e.g. for data like:
```
    "address": {
      "city": string,
      "street": string
    },
 ```
"Address" would not show up in the UI anymore.

The alternative solution would be to fix the only broken case to update the `label={title}` at https://github.com/robocorp/robocorp/blob/511cb5ce1af0fe84a9a94ce54b55c0a413183e1b/action_server/frontend/src/views/actions/components/ActionRun.tsx#L334

fabioz:
Updated PR UI:

image

rihardsgravis (approved):
👍

PR #373, opened 2024-04-16
docs: Fix documentation for encryption algo

perry-mitchell:
Docs mention `aes256-cdc`, which is a typo of `aes256-cbc`. This PR corrects it but changes it to `aes256-gcm` instead, which is a more-likely value for this use case.


perry-mitchell:
Thanks @fabioz ! :pray:

fabioz:
The algorithm used was actually `aes256-gcm` as you noted, but I did a subsequent fix because even it the code there were bad references (https://github.com/robocorp/robocorp/commit/369e64a4861bbe19343b20339c6baa8ce3aae08c).

Thanks for spotting it.

PR #423, opened 2024-07-04
Clearer error message during `truststore` bug

cmin764:
More context in [Linear](https://linear.app/robocorp/issue/DEV-756/robocorpvaulterror-robocorp-truststore-and-robocorp-library-4933)

How to test it (from @mmokko ):
> With [mitmproxy](https://www.notion.so/sema4ai/MITM-proxy-setup-how-to-b410ea556c0546df9978c551343ca952#ef3baf5dcd614c7db66a03b4a49d75c6) you can test the proxy setup, but having certain certs is a bit harder. Easiest is to use https://badssl.com/. From there you can get a [self signed cert](https://self-signed.badssl.com/) and add it to your system cert store and the package should find it from there automatically.

[Test Package](https://github.com/cmin764/robots/tree/main/test-truststore)


c02mo (approved):
LGTM

PR #425, opened 2024-07-05
Enable browser maximization

### LINKED ISSUE #260: Add `args` parameter to the `browser.configure` method from `robocorp-browser`
### Target components

browser

### Why & who

Custom `args` worked with `RPA.Browser.Playwright` and some users are surprised and frustrated that this is not supported in `configure`. The specific arg trying to be set is `--start-maximized`. The `viewport_size` parameter doesn't work for them or at least that is my understanding from discussing with the user.

### Feature or enhancement

In `RPA.Browser.Playwright` you could set custom args like so:

```
from RPA.Browser.Playwright import playwright

playwright.new_browser(args=['--start-maximized'])
```

But this parameter is not available in `configure` or `configure_context` within `robocorp-browser`.

### Solution & alternative

_No response_

cmin764:
Closes #260 

Start a maximized browser window by passing the `maximized=True` switch to the `configure()` function.


cmin764 (commented):
Self-reviewed and happy with the changes.

Potential improvements:
1. Ensure we don't override the `args` we set for the browser launch when configuring with `maximize`, but usually the user knows what it's doing when explicitly using `configure_launch()`, so that can be added there additionally when such construct is used.
2. Double-check if starting a maximized window requires and enforces `headless=False`, as currently we're strict about it by raising an error.

joshyorko:
close please

PR #435, opened 2025-01-06
UPDATE STARLETTE TO 0.40.0

akaday:


fabioz:
Hello @akaday, actually this code is deprecated now and is going to be removed soon (sorry, the README is misleading and will be updated to signal that).

The action server development now happens under https://github.com/Sema4AI/actions/

